PR #5, opened 2024-01-29
Remove initial API and replace with DpuOperatorConfig API

wizhaoredhat:


bn222:
Need to change api version to v1alpha1

bn222:
Thanks for clarifying the commit message. 

/lgtm

PR #7, opened 2024-02-02
Create new api module for external consumers

wizhaoredhat:
This creates a new "/api" go module with the existing API. This makes it possible for external projects/plugins to import the API structs for the DPU Operator without the larger set of dependencies in the project's top level go.mod.


wizhaoredhat:
> Can you rebase the pr, and clean up the commits? We could have the following commits:
> 
> 1. upgrade to golang 1.21 (and don't forget the docker files),
> 2. another to vendor stuff
> 3. Add empty api module that still passes on prow

As discussed offline, I have split the last PR into 2.

PR #11, opened 2024-02-21
Add CNI server GO test

wizhaoredhat:


wizhaoredhat:
@bn222 PTAL

dpu-cni/pkgs/cniserver/cniserver_test.go
```diff
@@ -0,0 +1,72 @@
...
+			os.Setenv("CNI_COMMAND", "ADD")
+			os.Setenv("CNI_CONTAINERID", cmdArgs.ContainerID)
+			os.Setenv("CNI_NETNS", cmdArgs.Netns)
+			os.Setenv("CNI_IFNAME", cmdArgs.IfName)
+			os.Setenv("CNI_PATH", cmdArgs.Path)
+			g.It("should get a correct response from the post request", func() {
+				resp, ver, err := plugin.PostRequest(cmdArgs)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				o.Expect(ver).To(o.Equal(cniVersion))
+				o.Expect(resp.Result).To(o.Equal(expectedResult))
```
bn222, line 68: What do we gain by having this test?
What does a passing test tell us?
If some commit starts to cause this to fail, what does that tell us about the config?

As per my understanding, `postRequest` is a translation from cni args to http request. It's a function that reads stuff and writes it somewhere else.

wizhaoredhat: What we gain: APIs should correctly parse the arguments. This should evolve as we use more of the arguments. This tests that any changes to the socket code won't affect anything. As this test, test both server and CNI shim side.
What a passing/failing test tells us:  As we do development we should really be aware to add more tests here to make sure relevant arguments are correctly expected on both the server and CNI shim side. The hope is that if we extend the arguments that we use the tests will start failing and the developer needs to add more test cases to make sure arguments are correctly passes between server and CNI shim. This is where `expectedResult` matters.
Yes `postRequest` comes from the CNI shim code side to send the CNI arguments over http.

PR #18, opened 2024-03-28
Add CniServer into HostDaemon

bn222:
This PR adds functionality to propagate a CNI add for a device that is added on the host down to the DPU which in turns forwards that to the vendor plugin on the DPU.

### DIFF
--- a/daemon/dpudaemon.go
+++ b/daemon/dpudaemon.go
@@ -36,28 +36,38 @@ func NewDpuDaemon(vsp plugin.VendorPlugin) *DpuDaemon {
 	}
 }
 
-func (d *DpuDaemon) Start() {
+func (d *DpuDaemon) Listen() (net.Listener, error) {
 	d.log.Info("starting DpuDaemon")
 	addr, port, err := d.vsp.Start()
 	if err != nil {
 		d.log.Error(err, "Failed to get addr:port from VendorPlugin")
 	}
-	d.server = grpc.NewServer()
-	pb.RegisterBridgePortServiceServer(d.server, d)
-
 	lis, err := net.Listen("tcp", fmt.Sprintf("%s:%d", addr, port))
 	if err != nil {
-		d.log.Error(err, "Failed to start listening")
+		d.log.Error(err, "Failed to start listening on", "addr", addr, "port", port)
+		return lis, err
 	}
+	d.log.Info("server listening", "address", lis.Addr())
 
-	go func() {
-		d.log.Info("server listening", "address", lis.Addr())
+	return lis, err
+}
 
-		if err := d.server.Serve(lis); err != nil {
-			d.log.Error(err, "Failed to start serving")
-			panic("Failed to listen")
-		}
-	}()
+func (d *DpuDaemon) ListenAndServe() error {
+	lis, err := d.Listen()
+	if err != nil {
+		return err
+	}
+	return d.Serve(lis)
+}
+
+func (d *DpuDaemon) Serve(listen net.Listener) error {
+	d.server = grpc.NewServer()
+	pb.RegisterBridgePortServiceServer(d.server, d)
+	if err := d.server.Serve(listen); err != nil {
+		d.log.Error(err, "Failed to start serving")
+		return err
+	}
+	return nil
 }
 
 func (d *DpuDaemon) Stop() {
--- a/daemon/hostdaemon.go
+++ b/daemon/hostdaemon.go
@@ -2,11 +2,17 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
+	"time"
 
+	cni100 "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/go-logr/logr"
 	"github.com/openshift/dpu-operator/daemon/plugin"
+	"github.com/openshift/dpu-operator/dpu-cni/pkgs/cniserver"
 	"github.com/openshift/dpu-operator/dpu-cni/pkgs/cnitypes"
+	"github.com/openshift/dpu-operator/dpu-cni/pkgs/sriov"
 	pb "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
@@ -22,17 +28,30 @@ type HostDaemon struct {
 	addr          string
 	port          int32
 	cniServerPath string
+	cniserver     *cniserver.Server
+	sm            sriov.Manager
 }
 
-func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int) (*pb.BridgePort, error) {
-	d.connectWithRetry()
+func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int, mac string) (*pb.BridgePort, error) {
+	err := d.connectWithRetry()
+	if err != nil {
+		return nil, err
+	}
+
+	m, err := net.ParseMAC(mac)
+	if err != nil {
+		return nil, err
+	}
+
 	createRequest := &pb.CreateBridgePortRequest{
 		BridgePort: &pb.BridgePort{
-			Name: fmt.Sprintf("%d-%d-%d", pf, vf, vlan),
+			Name: fmt.Sprintf("%d-%d", pf, vf),
 			Spec: &pb.BridgePortSpec{
-				Ptype:          1,
-				MacAddress:     []byte{},
-				LogicalBridges: []string{},
+				Ptype:      1,
+				MacAddress: m,
+				LogicalBridges: []string{
+					fmt.Sprintf("%d", vlan),
+				},
 			},
 		},
 	}
@@ -40,10 +59,10 @@ func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int) (*pb.BridgePort,
 	return d.client.CreateBridgePort(context.TODO(), createRequest)
 }
 
-func (d *HostDaemon) DeleteBridgePort(pf int, vf int, vlan int) error {
+func (d *HostDaemon) DeleteBridgePort(pf int, vf int, vlan int, mac string) error {
 	d.connectWithRetry()
 	req := &pb.DeleteBridgePortRequest{
-		Name: fmt.Sprintf("%d-%d-%d", pf, vf, vlan),
+		Name: fmt.Sprintf("%d-%d-%d-%s", pf, vf, vlan, mac),
 	}
 
 	_, err := d.client.DeleteBridgePort(context.TODO(), req)
@@ -55,12 +74,23 @@ func NewHostDaemon(vsp plugin.VendorPlugin) *HostDaemon {
 		vsp:           vsp,
 		log:           ctrl.Log.WithName("HostDaemon"),
 		cniServerPath: cnitypes.ServerSocketPath,
+		sm:            sriov.NewSriovManager(),
 	}
 }
 
-func (d *HostDaemon) connectWithRetry() {
+func (d *HostDaemon) WithCniServerPath(serverPath string) *HostDaemon {
+	d.cniServerPath = serverPath
+	return d
+}
+
+func (d *HostDaemon) WithSriovManager(manager sriov.Manager) *HostDaemon {
+	d.sm = manager
+	return d
+}
+
+func (d *HostDaemon) connectWithRetry() error {
 	if d.conn != nil {
-		return
+		return nil
 	}
 	// Might want to change waitForReady to true to
 	// block on connection. Currently, we connect
@@ -82,13 +112,51 @@ func (d *HostDaemon) connectWithRetry() {
 	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", d.addr, d.port), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithDefaultServiceConfig(retryPolicy))
 	if err != nil {
 		d.log.Error(err, "did not connect")
+		return err
 	}
 	d.conn = conn
 	d.client = pb.NewBridgePortServiceClient(conn)
+	return nil
 }
 
-func (d *HostDaemon) Start() {
-	d.log.Info("starting HostDaemon", "devflag", d.dev)
+func (d *HostDaemon) cniCmdAddHandler(req *cnitypes.PodRequest) (*cni100.Result, error) {
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.RuntimeConfig.Mac
+	d.log.Info("addHandler", "CNIConf", req.CNIConf)
+	vlan := *req.CNIConf.Vlan
+	d.log.Info("addHandler", "pf", pf, "vf", vf, "mac", mac, "vlan", vlan)
+	_, err := d.CreateBridgePort(pf, vf, vlan, mac)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to call CreateBridgePort: %v", err)
+	}
+	d.log.Info("addHandler CreateBridgePort succeeded")
+
+	res, err := d.sm.CmdAdd(req)
+	if err != nil {
+		return nil, errors.New("SRIOV manager failed in add handler")
+	}
+	d.log.Info("addHandler d.sm.CmdAdd succeeded")
+	return res, nil
+}
+
+func (d *HostDaemon) cniCmdDelHandler(req *cnitypes.PodRequest) (*cni100.Result, error) {
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.MAC
+	vlan := *req.CNIConf.Vlan
+	d.log.Info("delHandler", "pf", pf, "vf", vf, "mac", mac, "vlan", vlan)
+	d.DeleteBridgePort(pf, vf, vlan, mac)
+
+	err := d.sm.CmdDel(req)
+	if err != nil {
+		return nil, errors.New("SRIOV manager failed in del handler")
+	}
+	return nil, nil
+}
+
+func (d *HostDaemon) Listen() (net.Listener, error) {
+	d.log.Info("starting HostDaemon", "devflag", d.dev, "cniServerPath", d.cniServerPath)
 
 	addr, port, err := d.vsp.Start()
 	if err != nil {
@@ -96,4 +164,43 @@ func (d *HostDaemon) Start() {
 	}
 	d.addr = addr
 	d.port = port
+
+	add := func(r *cnitypes.PodRequest) (*cni100.Result, error) {
+		return d.cniCmdAddHandler(r)
+	}
+	del := func(r *cnitypes.PodRequest) (*cni100.Result, error) {
+		return d.cniCmdDelHandler(r)
+	}
+
+	d.cniserver = cniserver.NewCNIServer(add, del, cniserver.WithSocketPath(d.cniServerPath))
+
+	return d.cniserver.Listen()
+}
+
+func (d *HostDaemon) ListenAndServe() error {
+	listener, err := d.Listen()
+
+	if err != nil {
+		d.log.Error(err, "Failed to listen")
+		return err
+	}
+	return d.Serve(listener)
+}
+
+func (d *HostDaemon) Serve(listener net.Listener) error {
+	err := d.cniserver.Serve(listener)
+	if err != nil {
+		d.log.Error(err, "Error starting CNI server for shim")
+		return err
+	}
+	return nil
+}
+
+func (d *HostDaemon) Stop() {
+	if d.cniserver != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+		defer cancel()
+		d.cniserver.Shutdown(ctx)
+		d.cniserver = nil
+	}
 }
--- a/daemon/main.go
+++ b/daemon/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"io"
+	"net"
 	"os"
 
 	"github.com/openshift/dpu-operator/daemon/plugin"
@@ -16,7 +17,10 @@ import (
 var ()
 
 type Daemon interface {
-	Start()
+	Listen() (net.Listener, error)
+	ListenAndServe() error
+	Serve(listen net.Listener) error
+	Stop()
 }
 
 func isDpuMode(mode string) (bool, error) {
@@ -95,5 +99,5 @@ func main() {
 		log.Error(err, "Failed to start daemon")
 		return
 	}
-	daemon.Start()
+	daemon.ListenAndServe()
 }
--- a/daemon/main_test.go
+++ b/daemon/main_test.go
@@ -2,13 +2,21 @@ package main_test
 
 import (
 	"flag"
+	"os"
+	"path/filepath"
 
 	g "github.com/onsi/ginkgo/v2"
 	"go.uber.org/zap/zapcore"
+	"k8s.io/klog/v2"
 
 	. "github.com/onsi/gomega"
 
+	"github.com/containernetworking/cni/pkg/skel"
+	current "github.com/containernetworking/cni/pkg/types/100"
+	"github.com/containernetworking/plugins/pkg/ns"
 	. "github.com/openshift/dpu-operator/daemon"
+	"github.com/openshift/dpu-operator/dpu-cni/pkgs/cni"
+	"github.com/openshift/dpu-operator/dpu-cni/pkgs/cnitypes"
 	opi "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -38,6 +46,64 @@ func (v *DummyPlugin) DeleteBridgePort(deleteRequest *opi.DeleteBridgePortReques
 	return nil
 }
 
+type SriovManagerStub struct{}
+
+func (m SriovManagerStub) SetupVF(conf *cnitypes.NetConf, podifName string, netns ns.NetNS) error {
+	return nil
+}
+
+func (m SriovManagerStub) ReleaseVF(conf *cnitypes.NetConf, podifName string, netns ns.NetNS) error {
+	return nil
+}
+
+func (m SriovManagerStub) ResetVFConfig(conf *cnitypes.NetConf) error {
+	return nil
+}
+
+func (m SriovManagerStub) ApplyVFConfig(conf *cnitypes.NetConf) error {
+	return nil
+}
+
+func (m SriovManagerStub) FillOriginalVfInfo(conf *cnitypes.NetConf) error {
+	return nil
+}
+
+func (m SriovManagerStub) CmdAdd(req *cnitypes.PodRequest) (*current.Result, error) {
+	result := &current.Result{}
+	result.CNIVersion = req.CNIConf.CNIVersion
+	return result, nil
+}
+
+func (m SriovManagerStub) CmdDel(req *cnitypes.PodRequest) error {
+	return nil
+}
+
+func PrepArgs(cniVersion string, command string) *skel.CmdArgs {
+	cniConfig := "{\"cniVersion\": \"" + cniVersion + "\",\"name\": \"dpucni\",\"type\": \"dpucni\", \"RuntimeConfig\": {\"Mac\": \"00:11:22:33:44:55\"}, \"vlan\": 7}"
+	cmdArgs := &skel.CmdArgs{
+		ContainerID: "fakecontainerid",
+		Netns:       "fakenetns",
+		IfName:      "fakeeth0",
+		Args:        "",
+		Path:        "fakepath",
+		StdinData:   []byte(cniConfig),
+	}
+	os.Clearenv()
+	os.Setenv("CNI_COMMAND", command)
+	os.Setenv("CNI_ARGS", "K8S_POD_NAMESPACE=x;K8S_POD_NAME=y;K8S_POD_UID=z")
+	os.Setenv("CNI_CONTAINERID", cmdArgs.ContainerID)
+	os.Setenv("CNI_NETNS", cmdArgs.Netns)
+	os.Setenv("CNI_IFNAME", cmdArgs.IfName)
+	os.Setenv("CNI_PATH", cmdArgs.Path)
+
+	return cmdArgs
+}
+
+func cmdAdd(cniVersion string, serverSocketPath string) (*cnitypes.Response, string, error) {
+	p := &cni.Plugin{SocketPath: serverSocketPath}
+	return p.PostRequest(PrepArgs(cniVersion, cnitypes.CNIAdd))
+}
+
 var _ = g.BeforeSuite(func() {
 	opts := zap.Options{
 		Development: true,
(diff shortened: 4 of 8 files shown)


daemon/main.go
```diff
@@ -53,6 +80,12 @@ func main() {
 	log := ctrl.Log.WithName("Daemon Init")
 	log.Info("Daemon init")
 
+	err = copyCNI()
```
wizhaoredhat, line 83: Unfortunately due to the delta between RHEL versions we should be using the `entrypoint.sh` method to copy the CNI. Same code for SRIO-CNI and Multus:

https://github.com/openshift/sriov-cni/pull/100#issuecomment-2009001716

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	go.uber.org/zap v1.26.0
 	golang.org/x/net v0.19.0
 	google.golang.org/grpc v1.58.3
-	google.golang.org/protobuf v1.31.0
+	google.golang.org/protobuf v1.33.0
```
wizhaoredhat, line 24: Is this upgrade to utilize new APIs? Maybe a helpful commit message may make it easier to for people to know happened here.

daemon/hostdaemon.go
```diff
@@ -89,4 +99,20 @@ func (d *HostDaemon) Start() {
...
+	server := cniserver.NewCNIServer()
+	err = server.Start()
+
+	if err != nil {
+		d.log.Error(err, "Error starting CNI server for shim")
+	}
+
+	d.cniserver = server
+	go func() {	
+		d.cniserver.Start()
```
wizhaoredhat, line 168: This code is kind of confusing. We do server.Start() above and we are doing the same server.Start() in a Go routine.

daemon/main_test.go
```diff
@@ -27,18 +33,61 @@ var _ = g.BeforeSuite(func() {
...
 var _ = g.Describe("Main", func() {
 	g.Context("Connection", func() {
-		dummyPluginDPU  := plugin.NewDummyPlugin()
+		dummyPluginDPU := plugin.NewDummyPlugin()
 		DpuDaemon := NewDpuDaemon(dummyPluginDPU)
 
 		dummyPluginHost := plugin.NewDummyPlugin()
 		HostDaemon := NewHostDaemon(dummyPluginHost)
 
+		g.It("Should connect succesfully if the server is up first", func() {
```
wizhaoredhat, line 64: "Should connect successfully if the DPU Daemon is up first"

daemon/main_test.go
```diff
@@ -27,18 +33,61 @@ var _ = g.BeforeSuite(func() {
...
+			DpuDaemon.Start()
+			HostDaemon.Start()
+			err := HostDaemon.CreateBridgePort(1,1,1)
+			Expect(err).ShouldNot(HaveOccurred())
+
+			HostDaemon.Stop()
+			DpuDaemon.Stop()
+		})
 		g.It("Should connect succesfully if the server is up first", func() {
```
wizhaoredhat, line 73: "Should connect successfully if the Host Daemon is up first"

dpu-cni/pkgs/cniserver/cniserver.go
```diff
@@ -247,14 +248,23 @@ func (s *Server) handleCNIRequest(r *http.Request) ([]byte, error) {
...
 	if err != nil {
 		return nil, err
 	}
+	defer req.Cancel()
 
-	response := &cnitypes.Response{}
-	response.Result = result
+	var result *cni100.Result = nil
+	if req.Command == cnitypes.CNIAdd {
+		result, err = s.addHandler(req)
```
wizhaoredhat, line 262: Can we rename "addHandler" to  "cniCmdAddHandler"... "addHandler" and "delHandler" has different meanings on its own.

dpu-cni/pkgs/cniserver/cniserver_test.go
```diff
@@ -59,13 +42,35 @@ var _ = g.Describe("Cniserver", func() {
...
 	var err error
+	var addHandlerCalled bool = false
+	var delHandlerCalled bool = false
+
+	addHandler := func(request *cnitypes.PodRequest) (*current.Result, error) {
+		result := &current.Result{
+			CNIVersion: request.CNIConf.CNIVersion,
+		}
+
+		addHandlerCalled = true
```
wizhaoredhat, line 53: Let's rename "addHandler*" to something like "cniCmdAddHandler*"

dpu-cni/pkgs/cniserver/cniserver_test.go
```diff
@@ -80,11 +85,26 @@ var _ = g.Describe("Cniserver", func() {
 				CNIVersion: cniVersion,
 			}
 			g.It("should get a correct response from the post request", func() {
-				resp, ver, err := plugin.PostRequest(PrepArgs(cniVersion))
+				resp, ver, err := plugin.PostRequest(PrepArgs(cniVersion, "ADD"))
```
wizhaoredhat, line 86: (Optional) I know it is horrible on its on. But can we have a const added something for ADD and DEL? like "cnitypes.CommandAdd"

daemon/hostdaemon.go
```diff
@@ -26,11 +29,11 @@ type HostDaemon struct {
 	cniserver *cniserver.Server
 }
 
-func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int) (error) {
+func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int, mac string) (error) {
```
wizhaoredhat, line 33: (Comment) Remember how we might want to have a json for the name... We are fast approaching needing this.

daemon/hostdaemon.go
```diff
@@ -90,6 +93,36 @@ func (d *HostDaemon) ensureConnected() {
...
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.MAC
+	vlan := 7
+	d.CreateBridgePort(pf, vf, vlan, mac)
+
+	sm := sriov.NewSriovManager()
+	res, err := sm.CmdAdd(req)
+	if err != nil {
+	    return nil, errors.New("SRIOV manager falied in add handler")
```
wizhaoredhat: SP: Failed

daemon/hostdaemon.go
```diff
@@ -90,6 +93,36 @@ func (d *HostDaemon) ensureConnected() {
...
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.MAC
+	vlan := 7
+	d.DeleteBridgePort(pf, vf, vlan, mac)
+
+	sm := sriov.NewSriovManager()
+	err := sm.CmdDel(req)
+	if err != nil {
+	    return nil, errors.New("SRIOV manager falied in del handler")
```
wizhaoredhat, line 137: SP: Failed

wizhaoredhat (changes requested):
Part 1 of review stopping at commit "vendoring". Can we consolidate all the Go vendoring into 1 commit?

daemon/hostdaemon.go
```diff
@@ -51,21 +56,37 @@ func NewHostDaemon(vsp plugin.VendorPlugin) *HostDaemon {
 	}
 }
 
+func (d *HostDaemon) ensureConnected() {
+	if d.conn != nil {
```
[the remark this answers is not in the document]
bn222: When you `Dial` using `grpc`, the connection might not be up. For that reason, having an additional check here might not work. If there is no connection, it will error out later.

daemon/main.go
```diff
@@ -29,7 +29,7 @@ func isDpuMode(mode string) (bool, error) {
 }
 
 func createDaemon(dpuMode bool) (Daemon, error) {
-	plugin := plugin.NewGrpcPlugin(dpuMode)
+	plugin := plugin.NewDummyPlugin() // plugin.NewGrpcPlugin(dpuMode)
```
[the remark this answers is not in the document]
bn222: Thanks for catching this.

vendor/github.com/golang/protobuf/jsonpb/decode.go
Eoghan1232, line 1: can we remove vendor folder entirely?

bindata/daemon/04.daemonset.yaml
```diff
@@ -26,13 +26,38 @@ spec:
...
         env:
         - name: K8S_NODE
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
         - name: NAMESPACE
           value: {{.Namespace}}
+        volumeMounts:
+        - name: vendor-plugin-sock
+          mountPath: /var/run/dpu-daemon
```
Eoghan1232, line 39: /var/run/daemon/dpu-daemon?

daemon/dpudaemon.go
```diff
@@ -17,16 +17,17 @@ type DpuDaemon struct {
...
 	vsp plugin.VendorPlugin
 	log logr.Logger
+	server *grpc.Server
 }
 
 func (s *DpuDaemon) CreateBridgePort(context context.Context, bpr *pb.CreateBridgePortRequest) (*pb.BridgePort, error) {
-	s.log.Info("Create Bridge Port", "req", bpr.BridgePort.Name)
-	return &pb.BridgePort{}, nil
+	err := s.vsp.CreateBridgePort(bpr)
+	return nil, err
```
Eoghan1232, line 25: why ``nil, err``?

daemon/hostdaemon.go
```diff
@@ -58,14 +146,29 @@ func (d *HostDaemon) Start() {
...
+
+	add := func(r *cnitypes.PodRequest) (*cni100.Result, error) {
+		return d.addHandler(r)
+	}
+	del := func(r *cnitypes.PodRequest) (*cni100.Result, error) {
+		return d.delHandler(r)
+	}
+
+	server := cniserver.NewCNIServer(add, del)
+	err = server.Start()
```
Eoghan1232, line 160: why is this here?

daemon/hostdaemon.go
```diff
@@ -58,14 +146,29 @@ func (d *HostDaemon) Start() {
...
-	// listens to requests coming from the cni shim
-	select {}
+	d.cniserver = server
+	go func() {	
+		d.cniserver.Start()
+	}()
+}
+
+func (d *HostDaemon) Stop() {
+
```
Eoghan1232, line 173:
```suggestion
	if d.cniserver != nil {
		d.cniserver.GracefulStop()
		d.cniserver = nil
	}
```

daemon/main.go
```diff
@@ -37,6 +39,31 @@ func createDaemon(dpuMode bool) (Daemon, error) {
...
+	_, err = io.Copy(destinationFile, sourceFile)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func copyCNI() error {
+	return copyFile("/dpu-cni", "/opt/cni/bin/dpu-cni")
```
Eoghan1232, line 64: this should be part of an entrypoint script tbh, this is how sriov-cni does it
I don't think daemon should be responsible for copying the binary?

daemon/main.go
Eoghan1232: +1

daemon/plugin/vendorplugin.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	VendorPluginDir        string = "/var/run/daemon"
+	VendorPluginDir        string = "/var/run/daemon/vendor-plugin"
```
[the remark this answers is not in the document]
Eoghan1232: /var/run/daemon/dpu-daemon should be used

daemon/plugin/vendorplugin.go
```diff
@@ -3,50 +3,57 @@ package plugin
...
 	}
 }
 
-func (v *DummyPlugin) Start() (string, string) {
-	if runtime.GOARCH == "amd64" {
-		return "127.0.0.1", "50051"
-	} else {
-		return "127.0.0.1", "50051"
-	}
+func (v *DummyPlugin) Start() (string, int32, error) {
```
Eoghan1232, line 37: we can probably remove this now

daemon/plugin/vendorplugin.go
```diff
@@ -3,50 +3,57 @@ package plugin
...
+	return "127.0.0.1", 50051, nil
 }
 
 func (v *DummyPlugin) Stop() {
 
 }
 
+func (v *DummyPlugin) CreateBridgePort(createRequest *opi.CreateBridgePortRequest) error {
```
Eoghan1232, line 45: same comment as above

daemon/plugin/vendorplugin.go
```diff
@@ -59,6 +66,7 @@ func (g *GrpcPlugin) Start() (string, int32, error) {
 		g.log.Error(err, "Failed to start serving")
 		return "", 0, err
 	}
+	g.log.Info("Successful call to Init", ipPort.Ip, ipPort.Port)
```
Eoghan1232, line 69: this is just a debug line right? remove?

daemon/plugin/vendorplugin.go
```diff
@@ -94,5 +102,17 @@ func (g *GrpcPlugin) ensureConnected() error {
 	g.conn = conn
 
 	g.client = pb.NewLifeCycleServiceClient(conn)
+	g.opiClient = opi.NewBridgePortServiceClient(conn)
 	return nil
 }
+
+func (g *GrpcPlugin) CreateBridgePort(createRequest *opi.CreateBridgePortRequest) error {
+	g.ensureConnected()
+	_, err := g.opiClient.CreateBridgePort(context.TODO(), createRequest)
```
Eoghan1232, line 111: this is where we will forward req to vsp right?

dpu-cni/pkgs/cniserver/cniserver.go
```diff
@@ -24,17 +25,19 @@ import (
...
-	runDir         string
-	socketPath     string
+	addHandler processRequestFunc
+	delHandler processRequestFunc
+	socketPath string
 }
 
 // ensureRunDirExists makes sure that the socket being created is only accessible to root.
-func ensureRunDirExists(runDir string, socketPath string) error {
+func ensureRunDirExists(serverSocketPath string) error {
```
Eoghan1232, line 37: why is this needed?
on deletion should the socket not get cleaned up?

dpu-cni/pkgs/cniserver/cniserver.go
Eoghan1232: +1

dpu-cni/pkgs/cniserver/cniserver.go
```diff
@@ -319,15 +333,8 @@ func NewCNIServer(options ...func(*Server)) *Server {
...
 }
 
-func WithHandler(rqFunc processRequestFunc) func(*Server) {
-	return func(s *Server) {
-		s.processRequest = rqFunc
-	}
-}
-
-func WithSocketPath(runDir string, socketPath string) func(*Server) {
+func WithSocketPath(socketPath string) func(*Server) {
```
Eoghan1232, line 336: where is this func used?

Eoghan1232: also should the func be exported?

dpu-cni/pkgs/cnitypes/cnitypes.go
```diff
@@ -10,9 +10,7 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-const ServerRunDir string = "/var/run/dpu-cni"
-const ServerSocketName string = "dpu-cni-server.sock"
-const ServerSocketPath string = ServerRunDir + "/" + ServerSocketName
+const ServerSocketPath string = "/var/run/dpu-daemon/dpu-cni/dpu-cni-server.sock"
```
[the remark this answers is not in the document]
Eoghan1232: +1
```suggestion
const ServerSocketPath string = "/var/run/dpu-daemon/dpu-cni/dpu-cni-server.sock"
const path string = "/var/run/dpu-daemon/"
const path2 string = "dpu-cni/dpu-cni-server.sock"
```

with actual names but it's handy to have 2 separated so we can construct new paths in future without refactoring

dpu-cni/pkgs/sriov/sriov.go
```diff
@@ -198,9 +198,9 @@ func (s *sriovManager) ApplyVFConfig(conf *cnitypes.NetConf) error {
 		return fmt.Errorf("failed to lookup master %q: %v", conf.Master, err)
 	}
 	// 1. Set vlan
-	if err = s.nLink.LinkSetVfVlanQosProto(pfLink, conf.VFID, *conf.Vlan, *conf.VlanQoS, cnitypes.VlanProtoInt[*conf.VlanProto]); err != nil {
-		return fmt.Errorf("failed to set vf %d vlan configuration - id %d, qos %d and proto %s: %v", conf.VFID, *conf.Vlan, *conf.VlanQoS, *conf.VlanProto, err)
-	}
+	// if err = s.nLink.LinkSetVfVlanQosProto(pfLink, conf.VFID, *conf.Vlan, *conf.VlanQoS, cnitypes.VlanProtoInt[*conf.VlanProto]); err != nil {
```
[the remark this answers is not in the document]
Eoghan1232: no - this code is commented out right now due to driver not supporting this func.
I don't see a reason to create a flag based on support or not ? 

or you mean if another vendor supported and wanting this func  ?

dpu-cni/pkgs/cniserver/cniserver.go
bn222: `dpu-cni/pkgs/cniserver/cniserver_test.go`

daemon/plugin/vendorplugin.go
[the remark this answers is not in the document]
bn222: The intent is to have everything in `/var/run/dpu-daemon/`. The reason there is a subdirectory is that we need to set permissions on the directory.

For that reason, We should have:

`/var/run/dpu-daemon/vendor-plugin`
`/var/run/dpu-daemon/cni-server`
...

daemon/plugin/vendorplugin.go
bn222: I will move it to tests since that's where it's used and it's specific to testing.

daemon/main.go
bn222: Can you elaborate on why we need to do the copy outside of the daemon?

vendor/github.com/golang/protobuf/jsonpb/decode.go
bn222: We would like to but we can't.

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -40,6 +40,12 @@ type DpuOperatorConfigReconciler struct {
 //+kubebuilder:rbac:groups=config.openshift.io,resources=dpuoperatorconfigs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=config.openshift.io,resources=dpuoperatorconfigs/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=config.openshift.io,resources=dpuoperatorconfigs/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
bn222: I didn't run any operator-sdk commands. I added this to tell kubebuilder about what I was missing.

daemon/main.go
bn222: why not? What's the benefit of having an entrypoint script

dpu-cni/pkgs/sriov/sriov.go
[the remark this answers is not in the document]
bn222: I prefer to comment this out. It's a temporary workaround anyway.

daemon/plugin/vendorplugin.go
bn222: yes

daemon/plugin/vendorplugin.go
bn222: In case of CreateBridgePort in DummyPlugin (to be moved to test), I don't think we need to return an error. What would you want to return here?

Eoghan1232: *pb.BridgePort  ?

daemon/plugin/vendorplugin.go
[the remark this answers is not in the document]
bn222: What I propose is that we have `/var/run/dpu-daemon/` as a global var somewhere, and whenever we want to create a socket, we use concat it with `sub-dir/sock-name.socket`. We only need to pass around 1 variable. We can use `filepath.Dir` to get the directory and operate on that part separately.

daemon/main.go
bn222: Took a note that we will need to address this. More detail on this: cp from rhel 9 can be used on rhel 8, but we can't use our go binary if it is running on rhel 9 container while the host system is still rhel 8

bn222:
@SalDaniele PTAL

daemon/hostdaemon.go
bn222: Thanks for catching this. It's now resolved as part of the move to `ServeAndListen`.

daemon/plugin/vendorplugin.go
```diff
@@ -54,6 +54,7 @@ func (g *GrpcPlugin) Start() (string, int32, error) {
 		g.log.Error(err, "Failed to start serving")
 		return "", 0, err
 	}
+	g.log.Info("Successful call to Init", ipPort.Ip, ipPort.Port)
```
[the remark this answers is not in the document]
bn222: I've removed it. The assumption is that the vsp has proper tests

daemon/hostdaemon.go
```diff
@@ -31,25 +32,33 @@ type HostDaemon struct {
...
-	_, err := d.client.CreateBridgePort(context.TODO(), createRequest)
+	_, err = d.client.CreateBridgePort(context.TODO(), createRequest)
 	return err
 }
 
 func (d *HostDaemon) DeleteBridgePort(pf int, vf int, vlan int, mac string) (error) {
 	d.ensureConnected()
 	req := &pb.DeleteBridgePortRequest{
-		    Name: fmt.Sprintf("%d-%d-%d-%s", pf, vf, vlan, mac),
+		Name: fmt.Sprintf("%d-%d-%d-%s", pf, vf, vlan, mac),
```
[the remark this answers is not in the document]
bn222: The call to DeleteBridgePort isn't the same. I'm marking it as resolved since we don't need to change this right now but we might want to revisit this.

dpu-cni/pkgs/cniserver/cniserver.go
bn222: No, we must ensure that we are dealing with a clean slate. You can't assume anything about the state of the filesystem.

dpu-cni/pkgs/cnitypes/cnitypes.go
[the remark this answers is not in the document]
bn222: see my other comment

go.mod
vrindle: Any chance we can split this up upgrades into seperate PRs? These changes to go.mod and go.sum make a lot of additional noise in the PR.

daemon/dpudaemon.go
vrindle: I have the same question. I have the same question, why are no longer returning an empty bridge port and instead of nil and err?  If we are only going to return nil, it would make sense to return only err.

daemon/dpudaemon.go
```diff
@@ -21,13 +21,13 @@ type DpuDaemon struct {
...
 }
 
 func (s *DpuDaemon) DeleteBridgePort(context context.Context, bpr *pb.DeleteBridgePortRequest) (*emptypb.Empty, error) {
-	s.log.Info("Delete Bridge Port", "req", bpr.Name)
-	return nil, nil
+	err := s.vsp.DeleteBridgePort(bpr)
+	return nil, err
```
vrindle, line 30: same comment as above

daemon/hostdaemon.go
```diff
@@ -88,6 +107,43 @@ func (d *HostDaemon) connectWithRetry() {
 	d.client = pb.NewBridgePortServiceClient(conn)
 }
 
+func (d *HostDaemon) addHandler(req *cnitypes.PodRequest) (*cni100.Result, error) {
+	pf := 0
```
vrindle, line 111: We don't want magic numbers in the codebase. Is it possible to make these as constants?

daemon/hostdaemon.go
```diff
@@ -88,6 +107,43 @@ func (d *HostDaemon) connectWithRetry() {
...
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.MAC
+	vlan := 7
```
vrindle, line 114: Same as comment above

daemon/hostdaemon.go
```diff
@@ -88,6 +107,43 @@ func (d *HostDaemon) connectWithRetry() {
...
+	}
+	d.log.Info("addHandler sm.CmdAdd succeeded")
+	return res, nil
+}
+
+func (d *HostDaemon) delHandler(req *cnitypes.PodRequest) (*cni100.Result, error) {
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.MAC
+	vlan := 7
```
vrindle, line 135: We are reusing these. They should definitely be in a const.

daemon/hostdaemon.go
Eoghan1232: these are dummy values for testing and will be removed.
the actual values will come from the req.

daemon/hostdaemon.go
Eoghan1232: same as above.

@bn222 can you remove these and use values from the req instead?

Eoghan1232:
daemon/dpudaemon.go
I needed to add a `select{}` at the end of the Start() func - so the daemon would not exit out and complete afterwards

@bn222 we know about this, but it never got added to the PR I guess.

daemon/main_test.go
```diff
@@ -59,7 +59,7 @@ var _ = g.Describe("Main", func() {
...
 			HostDaemon.Start()
 			DpuDaemon.Start()
-			_, err := HostDaemon.CreateBridgePort(1, 1, 1)
+			_, err := HostDaemon.CreateBridgePort(1, 1, 1, "00:11:22:33:44:55")
```
wizhaoredhat: Squash this with the previous commit.

daemon/hostdaemon.go
```diff
@@ -34,23 +34,36 @@ type HostDaemon struct {
...
-			Spec: &pb.BridgePortSpec{
-				Ptype:      1,
-				MacAddress: macBytes,
-				LogicalBridges: []string{
-					fmt.Sprintf("%d", vlan),
-				},
-			},
-		},
-	}
+    BridgePort: &pb.BridgePort{
```
wizhaoredhat: Indentation issues.

daemon/hostdaemon.go
```diff
@@ -34,23 +34,36 @@ type HostDaemon struct {
...
+        Name: "m" + fmt.Sprint(1),
```
wizhaoredhat: I am confused what this is...

daemon/hostdaemon.go
```diff
@@ -34,23 +34,36 @@ type HostDaemon struct {
...
+        Spec: &pb.BridgePortSpec{
+            Ptype:          1,
+            MacAddress:     m,
+            LogicalBridges: []string{fmt.Sprint(1 + 2)},
```
wizhaoredhat: Missing VLAN and PF/VLAN info???

wizhaoredhat: What is wrong with using the commented out code from below?

daemon/hostdaemon.go
```diff
@@ -161,7 +174,7 @@ func (d *HostDaemon) Start() {
 		return d.delHandler(r)
 	}
 
-	d.cniserver = cniserver.NewCNIServer(add, del)
+	d.cniserver = cniserver.NewCNIServer(add, del, cniserver.WithSocketPath(d.cniServerPath))
```
wizhaoredhat, line 188: I think this code change should be squash with the first commit in this series for "each commit" should compile.

daemon/hostdaemon.go
```diff
@@ -63,7 +67,6 @@ func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int, mac string) (*pb
 //			},
 //		},
 //	}
-	print("------")
```
wizhaoredhat: squash with previous commit

daemon/main_test.go
```diff
@@ -45,6 +46,36 @@ func (v *DummyPlugin) DeleteBridgePort(deleteRequest *opi.DeleteBridgePortReques
 	return nil
 }
 
+type SriovManagerMock struct {}
```
wizhaoredhat, line 49: Just a general question whether we will be using Mocks for HW functionality?

wizhaoredhat (changes requested):

daemon/main_test.go
bn222: We should use a stub (or ideally a fake) when dealing with deeper layers of the architecture. We shouldn't blindly mock everything that goes to the hardware. Rather, we should write code that encapsulates behavior.

daemon/main_test.go
```diff
@@ -49,18 +115,58 @@ var _ = g.BeforeSuite(func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			go func() {
+				err = dpuDaemon.Serve(dpuListen)
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			hostListen, err := hostDaemon.Listen()
+			Expect(err).NotTo(HaveOccurred())
+			go func() {
+				_ = hostDaemon.Serve(hostListen)
```
wizhaoredhat, line 158: Same as above we should check the return `err`?

daemon/main_test.go
```diff
@@ -49,18 +115,58 @@ var _ = g.BeforeSuite(func() {
...
+		hostDaemon.Stop()
+	})
 
-		g.It("Should connect succesfully if the server is up first", func() {
-			HostDaemon.Start()
-			DpuDaemon.Start()
-			_, err := HostDaemon.CreateBridgePort(1, 1, 1)
-			Expect(err).ShouldNot(HaveOccurred())
+	g.Context("Connection", func() {
+		g.It("Should connect successfully if the DPU Daemon is up first", func() {
```
wizhaoredhat, line 147: Shouldn't we have another test case to see if Host Daemon is started first?

daemon/hostdaemon.go
```diff
@@ -82,18 +112,96 @@ func (d *HostDaemon) connectWithRetry() {
...
+
+	if err != nil {
+		d.log.Error(err, "Failed to listen")
+		return err
+	}
+	return d.Serve(listener)
+}
+
+func (d *HostDaemon) Serve(listner net.Listener) error {
+	err := d.cniserver.Serve(listner)
```
wizhaoredhat, line 192: SP: listener

daemon/hostdaemon.go
```diff
@@ -82,18 +112,96 @@ func (d *HostDaemon) connectWithRetry() {
...
+	}
+	d.log.Info("addHandler d.sm.CmdAdd succeeded")
+	return res, nil
+}
+
+func (d *HostDaemon) cniCmdDelHandler(req *cnitypes.PodRequest) (*cni100.Result, error) {
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.MAC
+	vlan := 7//*req.CNIConf.Vlan
```
wizhaoredhat, line 148: We should just use req.CNIConf.Vlan here. I don't see why this won't work.

daemon/hostdaemon.go
```diff
@@ -82,18 +112,96 @@ func (d *HostDaemon) connectWithRetry() {
...
 	d.conn = conn
 	d.client = pb.NewBridgePortServiceClient(conn)
+	return nil
+}
+
+func (d *HostDaemon) cniCmdAddHandler(req *cnitypes.PodRequest) (*cni100.Result, error) {
+	pf := 0
+	vf := req.CNIConf.VFID
+	mac := req.CNIConf.RuntimeConfig.Mac
+	d.log.Info("addHandler", "CNIConf", req.CNIConf)
```
wizhaoredhat, line 126: I think we can remove these prints.

wizhaoredhat (changes requested):
Overall looks good after squashing.

daemon/main_test.go
bn222: We should not because serve can fail and that's expected.

daemon/main_test.go
bn222: No, not today. In the future, we might want to abstract out using a robust way to communicate. It shouldn't be tested at this level imo.

daemon/hostdaemon.go
bn222: thanks for catching this commented out code.

daemon/hostdaemon.go
bn222: Are you sure? last time I removed logs, the first thing we ended up doing during debug sessions is put the logs back.

daemon/main_test.go
wizhaoredhat: But serve is not allowed to fail for the DPU Daemon?

daemon/main_test.go
wizhaoredhat: What is the purpose of the test, just to make sure the DPU/Host Daemon can talk to each other?

daemon/hostdaemon.go
wizhaoredhat: I think the VLAN one is redundant. But very true for CNIConf

daemon/main_test.go
bn222: This tests the fact that CNI calls get processed correctly. In the future, we will inject a dependency into HostDaemon to test the cases where we do plumbing.

bn222:
@wizhaoredhat ptal

PR #20, opened 2024-03-29
Fix daemon deployment from operator

bn222:
Some of the RBAC permissions were not properly set, which precluded the daemonset from being created on a fresh cluster.

In addition, also fix the version of a tool used in the Makefile to avoid randomly breaking builds.


bn222:
@SalDaniele PTAL

PR #21, opened 2024-03-29
Fix failing CNI server tests

bn222:


bn222:
@SalDaniele PTAL this first

SalDaniele:
LGTM

PR #23, opened 2024-03-30
Move server into go routine and add test

bn222:


bn222:
@SalDaniele @wizhaoredhat PTAL

daemon/hostdaemon.go
```diff
@@ -51,21 +56,37 @@ func NewHostDaemon(vsp plugin.VendorPlugin) *HostDaemon {
 	}
 }
 
+func (d *HostDaemon) ensureConnected() {
+	if d.conn != nil {
+		return
+	}
+	retryPolicy := `{
+		"methodConfig": [{
+		  "waitForReady": false,
```
wizhaoredhat, line 65: Add a comment about waitForReady. It is not obvious what this does. We should say the effect is this makes Dial nonblocking and returns immediately.

daemon/daemon_test.go
```diff
@@ -0,0 +1,13 @@
+package main_test
```
wizhaoredhat, line 1: Why do we need a main test? Why can't this test be called vendorplugin_test?

dpu-cni/pkgs/cniserver/cniserver_test.go
```diff
@@ -35,6 +35,26 @@ func processRequest(request *cnitypes.Request) (*current.Result, error) {
...
+	cmdArgs := &skel.CmdArgs{
+		ContainerID: "fakecontainerid",
+		Netns:       "fakenetns",
+		IfName:      "fakeeth0",
+		Args:        "",
+		Path:        "fakepath",
+		StdinData:   []byte(cniConfig),
+	}
+	os.Clearenv()
+	os.Setenv("CNI_COMMAND", "ADD")
```
wizhaoredhat, line 49: We should add macros for ADD and DEL and CHECK

daemon/daemon_test.go
bn222: We're actually testing the daemon, and since daemon is in main, I've used main_test.

bn222:
I've addressed all your concerns in separate commits. I will squash them when you have reviewed the changes.

PR #25, opened 2024-04-01
Add make local-deploy

bn222:
Useful for developing and deploying locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ endif
 # Be aware that the target commands are only tested with Docker which is
 # scaffolded by default. However, you might want to replace it to use other
 # tools. (i.e. podman)
-CONTAINER_TOOL ?= docker
+CONTAINER_TOOL ?= podman
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,12 @@ endif
 # tools. (i.e. podman)
 CONTAINER_TOOL ?= podman
 
+
+# Use the image urls from the yaml that is used with Kustomize for local
+# development.
+DPU_OPERATOR_IMAGE := $(shell yq -r .spec.template.spec.containers[0].image config/dev/local-images.yaml)
+DPU_DAEMON_IMAGE := $(shell yq -r .spec.template.spec.containers[0].env[0].value config/dev/local-images.yaml)
+
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
@@ -136,6 +142,18 @@ docker-build: test ## Build docker image with the manager.
 docker-push: ## Push docker image with the manager.
 	$(CONTAINER_TOOL) push ${IMG}
 
+GO_CONTAINER_CACHE = /tmp/dpu-operator-cache
+
+.PHONY: images-build
+images-build: ## Build all container images necessary to run the whole operator
+	mkdir -p $(GO_CONTAINER_CACHE)
+	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
+	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.daemon.rhel -t $(DPU_DAEMON_IMAGE)
+
+.PHONY: images-push
+images-push: test ## Push all container images necessary to run the whole operator
+	$(CONTAINER_TOOL) push $(DPU_OPERATOR_IMAGE)
+	$(CONTAINER_TOOL) push $(DPU_DAEMON_IMAGE)
 # PLATFORMS defines the target platforms for  the manager image be build to provide support to multiple
 # architectures. (i.e. make docker-buildx IMG=myregistry/mypoperator:0.0.1). To use this option you need to:
 # - able to use docker buildx . More info: https://docs.docker.com/build/buildx/
@@ -171,6 +189,10 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+
+.PHONY: local-deploy
+local-deploy: manifests kustomize ## Deploy controller with images hosted on local registry
+	$(KUSTOMIZE) build config/dev | $(KUBECTL) apply -f -
 
 .PHONY: undeploy
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -136,6 +136,15 @@ docker-build: test ## Build docker image with the manager.
 docker-push: ## Push docker image with the manager.
```
vrindle, line 136: Unrelated to this PR we should consider finding a way to build and push images outside of a Makefile

config/dev/local-images.yaml
```diff
@@ -0,0 +1,26 @@
...
+    app.kubernetes.io/managed-by: kustomize
+spec:
+  template:
+    spec:
+      containers:
+      - command:
+        name: manager
+        env:
+        - name: DPU_DAEMON_IMAGE
+          value: localhost:5000/dpu-daemon
```
vrindle, line 22: Can we change this to set tag the version explicitly? Docker images are meant to be immutable. Any time we use an untagged image, we are not sure which image is actually being run. We especially don't want this to be used for production workloads.

config/dev/local-images.yaml
```diff
@@ -0,0 +1,26 @@
...
+        - name: IMAGE_PULL_POLICIES
+          value: Always
+        image: localhost:5000/dpu-operator
```
vrindle, line 25: same as comment above

Makefile
```diff
@@ -136,6 +136,15 @@ docker-build: test ## Build docker image with the manager.
...
 	$(CONTAINER_TOOL) push ${IMG}
 
+.PHONY: images-build
+images: test ## Build all container images
```
Eoghan1232, line 140: I would add ``ifneq (, $(IMAGE_REGISTRY))`` to the top of the Makefile
so then you can construct the image name later on.
```
IMAGE_TAG_BASE = $(IMAGE_REGISTRY)/$(IMAGE_NAME)
IMAGE_TAG_LATEST?=$(IMAGE_TAG_BASE):latest
```

etc.

config/dev/local-images.yaml
```diff
@@ -0,0 +1,26 @@
...
+      containers:
+      - command:
+        name: manager
+        env:
+        - name: DPU_DAEMON_IMAGE
+          value: localhost:5000/dpu-daemon:dev
+        - name: IMAGE_PULL_POLICIES
+          value: Always
+        image: localhost:5000/dpu-operator:dev
+        imagePullPolicy: Always
```
vrindle, line 26: @bn222 Why are we setting `imagePullPolicy: Always`? If the image is meant to be immutable (which it seems like it is based on the tag above) it might make more sense to do `imagePullPolicy: IfNotPresent`?

Makefile
```diff
@@ -136,6 +136,17 @@ docker-build: test ## Build docker image with the manager.
...
+.PHONY: images-build
+images-build: test ## Build all container images
+	DPU_OPERATOR_IMAGE := $(shell yq -r .spec.template.spec.containers[0].image config/dev/local-images.yaml)
+	DPU_DAEMON_IMAGE := $(shell yq -r .spec.template.spec.containers[0].env[0].value config/dev/local-images.yaml)
+	$(CONTAINER_TOOL) build -v /tmp/go:/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
+	$(CONTAINER_TOOL) build -v /tmp/go:/go:z -f Dockerfile.daemon.rhel $(DPU_DAEMON_IMAGE)
+
+.PHONY: images-push
+images-push: test ## Build all container images
+	$(CONTAINER_TOOL) push localhost:5000/dpu-operator
```
Eoghan1232, line 148: the image names should come from the spec as well?

bn222: done

config/dev/local-images.yaml
bn222: when you build, you push to this image, then you deploy it with `make local-deploy`. If you overwrite the images, you want the latest. This patch is meant to make developer's life easy. That's why it also enables `imagePullPolicy: Always`

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -76,16 +77,29 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	return ctrl.Result{}, nil
 }
 
+
+func getImagePullPolicy() string {
+	if os.Getenv("IMAGE_PULL_POLICIES") == "" {
+		return "IfNotPresent"
+	} else {
+		return "Always"
```
Eoghan1232, line 85: doesn't this mean no matter what, you will override the pull policy?
for example, if you have env var set to ``IfNotPresent``, then it would return ``Always`` since the check only confirms if the var is empty, otherwise return ``Always``

Makefile
```diff
@@ -136,6 +142,15 @@ docker-build: test ## Build docker image with the manager.
 docker-push: ## Push docker image with the manager.
 	$(CONTAINER_TOOL) push ${IMG}
 
+.PHONY: images-build
+images-build: test ## Build all container images
+	$(CONTAINER_TOOL) build -v /tmp/go:/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
+	$(CONTAINER_TOOL) build -v /tmp/go:/go:z -f Dockerfile.daemon.rhel $(DPU_DAEMON_IMAGE)
```
Eoghan1232, line 148:
```suggestion
	$(CONTAINER_TOOL) build  -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
	$(CONTAINER_TOOL) build -f Dockerfile.daemon.rhel -t $(DPU_DAEMON_IMAGE)
```

bindata/daemon/04.daemonset.yaml
```diff
@@ -26,6 +26,7 @@ spec:
         image: {{.DpuOperatorDaemonImage}}
```
Eoghan1232, line 26: this image no longer gets picked up?

I had to hard code the value for local testing...
i.e.
``{{.DpuOperatorDaemonImage}}`` -> ``localhost:5001/dpu-daemon:dev``

internal/controller/dpuoperatorconfig_controller.go
bn222: good catch thanks.

bn222:
@Eoghan1232 @SalDaniele @wizhaoredhat PTAL

Makefile
```diff
@@ -68,6 +68,12 @@ endif
 # tools. (i.e. podman)
 CONTAINER_TOOL ?= podman
 
+
+# Use the image urls from the yaml that is used with Kustomize for local
+# development.
+DPU_OPERATOR_IMAGE := $(shell yq -r .spec.template.spec.containers[0].image config/dev/local-images.yaml)
```
wizhaoredhat, line 74: Conventions is to wrap yq to $(yq) and add a make rule to install yq. However I am ok with this.

Makefile
```diff
@@ -136,6 +142,18 @@ docker-build: test ## Build docker image with the manager.
...
+GO_CONTAINER_CACHE = /tmp/dpu-operator-cache
+
+.PHONY: images-build
+images-build: test ## Build all container images
+	mkdir -p $(GO_CONTAINER_CACHE)
+	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
+	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.daemon.rhel -t $(DPU_DAEMON_IMAGE)
+
+.PHONY: images-push
+images-push: test ## Build all container images
```
wizhaoredhat, line 154: Comment is not accurate.

You might want to add a small commentary on how to use this.

Create Local Registry.
make images-build
make images-push 
make local-deploy
Expect deployment to be created with DPU operator and DPU daemon

bn222: Doesn't this build all images necessary to run the whole operator?

bn222: @wizhaoredhat PTAL

PR #33, opened 2024-04-11
Fix mounting cni dir and sriov manager add call

bn222:


bindata/daemon/04.daemonset.yaml
```diff
@@ -39,7 +39,7 @@ spec:
           mountPath: /var/run/dpu-daemon
           mountPropagation: Bidirectional
         - name: cni-dir
-          mountPath: /opt/cni/bin
+          mountPath: /var/lib/cni
```
wizhaoredhat, line 42: We need to revisit this in the future. Because vanilla k8s uses /opt/cni/bin but OCP decided it was a good idea to use /var/lib/cni (probably because of COREOS). Can you put a "TODO" such that we can keep track of this and fix it in the future when we move this operator upstream?

PR #34, opened 2024-04-11
Add Create/Delete NF API

bn222:


wizhaoredhat:
Before letting this PR, we need to talk with DPU vendor(s) creating the VSP to agree to this API.

vrindle:
For the record the e2e tests are gettting triggered. They are failing because the go version is off so I have to fix that. @bn222 @wizhaoredhat @SalDaniele

dpu-api/api.proto
```diff
@@ -16,3 +21,10 @@ message IpPort {
   string ip = 1;
   int32 port = 2;
 }
+
+message NFRequest {
+  string input = 1;
```
wizhaoredhat, line 26: input_port/output_port?

wizhaoredhat: Fine as it is right now. Will revist in the future for better naming.

PR #35, opened 2024-04-11
Remove dependency on os dpuoperatorconfig controller

bn222:


bn222:
This depends on https://github.com/openshift/dpu-operator/pull/61

internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -156,9 +156,13 @@ func startDPUControllerManager(ctx context.Context, client *rest.Config, wg *syn
 		LeaderElectionID: "1e46962d.openshift.io",
 	})
 	Expect(err).NotTo(HaveOccurred())
+	cwd, err := os.Getwd()
```
wizhaoredhat, line 158: I don't see how much we gain here when we have to have an OS call here to get the root directory.

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -36,8 +36,39 @@ import (
...
 type DpuOperatorConfigReconciler struct {
 	client.Client
-	Scheme      *runtime.Scheme
-	BinDataPath string
+	Scheme          *runtime.Scheme
+	BinDataPath     string
+	imagePullPolicy string
+}
+
+type DpuOperatorConfigReconcilerBuilder struct {
```
wizhaoredhat, line 44: What do we gain with this new infrastructure?  We now have 2 objects with the same variables...

internal/controller/dpuoperatorconfig_controller_test.go
bn222: The main gain is that this PR removes the `os` as a dependency in the reconciler. That means that for the sake of readability, you will know that there is no use of env (i.e. no hidden input on that front) to the reconciler anywhere.

I agree that the test now depends on being run from the right directory. I will work on removing that.

internal/controller/dpuoperatorconfig_controller.go
bn222: Good point. I see how the main intent of this code got lost and how this is actually worst. I will remove it.

internal/controller/bindata/daemon/01.daemon_sa.yaml
wizhaoredhat, line 1: Why do we need to move the files?

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -18,24 +18,48 @@ package controller
...
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+//go:embed bindata/*
```
wizhaoredhat, line 40: I guess this can't reference the root directory?

bn222: Correct. It works relatively to the `.go` file, which is nice because as long as the `.go` file lives in the same directory as the embedded files, this will work.

internal/controller/bindata/daemon/01.daemon_sa.yaml
bn222: Because:
1. The files are not relevant top level2
2. The embedding works relatively to the embed line.

PR #37, opened 2024-04-12
Fix path and move call to sm earlier

bn222:
sm.CmdAdd has a side effect. It sets fields in its arg.


bn222:
This helped us to move past a few errors.

PR #39, opened 2024-04-12
Add examples directory with two DPUOperatorConfig CRs

bn222:


examples/dpu.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: config.openshift.io/v1
+kind: DpuOperatorConfig
+metadata:
+  name: dpu-operator-config
+  namespace: dpu-operator-system
```
wizhaoredhat, line 5: Q1) is this the namespace we will be using for the operator?

Q2) we probably want to have a yaml to create the aforementioned namespace.

bn222: Q1) short term yes

Q2) in the future, I agree.

examples/dpu.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: config.openshift.io/v1
```
wizhaoredhat, line 1: ---
apiVersion: v1
kind: Namespace
metadata:
  name: dpu-operator-system
--- 
<rest of yaml for each file>

wizhaoredhat: dpu-operator-system already created by operator sdk

PR #40, opened 2024-04-12
Add device plugin api

bn222:


dpu-api/api.proto
```diff
@@ -28,3 +28,21 @@ message NFRequest {
...
+
+service DeviceService {
+  rpc GetDevices(Empty) returns (DeviceListResponse);
+}
+
+message TopologyInfo {
+  string node = 1;
+}
+
+message Device {
```
wizhaoredhat, line 40: This is Eoghan's API right? I think it is proper to add his attribution.

PR #46, opened 2024-04-16
Fix Device plugin API arguments

bn222:


dpu-cni/pkgs/sriov/sriov.go
```diff
@@ -567,6 +567,7 @@ func (sm *sriovManager) CmdDel(req *cnitypes.PodRequest) error {
 			return err
 		}
 	}
+	req.CNIConf.VFID = netConf.VFID
```
wizhaoredhat, line 570: Why do we need this?

bn222: Because the VFID is stored in a cache and CmdDel didn't store the vfid in the req.

daemon/hostdaemon.go
```diff
@@ -45,12 +45,13 @@ func (d *HostDaemon) CreateBridgePort(pf int, vf int, vlan int, mac string) (*pb
 
 	createRequest := &pb.CreateBridgePortRequest{
 		BridgePort: &pb.BridgePort{
-			Name: fmt.Sprintf("%d-%d", pf, vf),
+			Name: "host" + fmt.Sprintf("%d-%d", pf, vf),
```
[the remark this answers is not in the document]
bn222: Because we need to start interface names with a string

dpu-cni/pkgs/sriov/sriov.go
wizhaoredhat: What does "req.CNIConf.VFID " give you prior to this change?

wizhaoredhat: Is perhaps,  LoadConfFromCache not working properly?

daemon/hostdaemon.go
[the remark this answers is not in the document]
wizhaoredhat: ok

dpu-cni/pkgs/sriov/sriov.go
bn222: it was `0`. LoadConfFromCache works well, but it loads `netConf`

wizhaoredhat: Ah and the code uses "netConf" in all its functions like "ResetVFConfig()"
But since we can only consume "req.CNIConf" outside of the "cmdAdd()" it doesn't work

wizhaoredhat: We would want a function called "fillInCNIConf" in cmdDel since this won't be the last place we need it.

PR #47, opened 2024-04-16
Add Openshift ART team specific requirements

wizhaoredhat:


Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 4.16.0
```
SalDaniele, line 6: Trying to understand why we are setting this to 4.16.0? Just to match the OCP release? I didn't see an equivalent field set on SNO

wizhaoredhat: https://github.com/openshift/sriov-network-operator/blob/e6d4576155f0cb427c37a8b276ebca889a830d1e/Makefile.bundle#L3 
Yes to match the OCP release version.

PR #57, opened 2024-04-26
Change the builder to RHEL9

wizhaoredhat:
The ART team has issues building with RHEL8. Thus we need to upgrade to RHEL9.


wizhaoredhat:
@bn222 / @SalDaniele PTAL

PR #58, opened 2024-04-26
Let errors occur when copying the DPU CNI

wizhaoredhat:
It is most likely that we will in the future include a command line argument for the CNI path since different cluster types will have different CNI locations.

For now remove the check for "writeable()" and simplify the code to just attempt to copy and make executable. Errors are logged and not act upon.
    
The writeable() code also had a bug where it would check for the file and not the directory.

### DIFF
--- a/daemon/main.go
+++ b/daemon/main.go
@@ -10,7 +10,6 @@ import (
 	deviceplugin "github.com/openshift/dpu-operator/daemon/device-plugin"
 	"github.com/openshift/dpu-operator/daemon/plugin"
 	"go.uber.org/zap/zapcore"
-	"golang.org/x/sys/unix"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
```
2024-04-29T13:54:25Z    INFO    Daemon Init     Daemon init
2024-04-29T13:54:25Z    ERROR   Daemon Init     Failed to prepare CNI binary in /opt    {"error": "open /opt/cni/bin/dpu-cni: read-only file system"}
main.main
        /workspace/daemon/main.go:118
runtime.main
        /usr/lib/golang/src/runtime/proc.go:267
2024-04-29T13:54:25Z    INFO    HostDaemon      starting HostDaemon     {"devflag": false, "cniServerPath": "/var/run/dpu-daemon/dpu-cni/dpu-cni-server.sock"}
I0429 13:54:25.990135  788336 cniserver.go:320] DPU CNI Server creating new router.
I0429 13:54:25.990169  788336 cniserver.go:40] Removing /var/run/dpu-daemon/dpu-cni
I0429 13:54:25.990224  788336 cniserver.go:65] Creating /var/run/dpu-daemon/dpu-cni
I0429 13:54:25.990326  788336 cniserver.go:82] Listen on socket path: /var/run/dpu-daemon/dpu-cni/dpu-cni-server.sock
```
Logs look like this on a CoreOS based deployment.

PR #59, opened 2024-05-01
Update controller-tools

bn222:


bn222:
closing this while we can't use go 1.22 yet.

PR #61, opened 2024-05-02
Testability

bn222:
Use a kind cluster instead of only a control plane for integration tests. This reduces the run time of tests from 42 seconds to 2 seconds on my system for subsequent runs since the cluster is not destroyed.


internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -1,6 +1,9 @@
 package controller
 
 import (
+	// "context"
```
wizhaoredhat, line 4: We should just remove these commented out imports.

internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -30,25 +53,14 @@ var (
 	testDpuDaemonName         = "dpu-daemon"
 	testSriovDevicePlugin     = "sriov-device-plugin"
 	testNetworkFunctionNAD    = "dpunfcni-conf"
-	testAPITimeout            = time.Second * 20
-	testRetryInterval         = time.Second * 1
+	testClusterName           = "dpu-operator-test-cluster"
+	testAPITimeout            = time.Second * 5
+	testRetryInterval         = time.Millisecond * 10
```
wizhaoredhat, line 58: On average, do you know how long it takes for an k8s API takes on kind? This seems a bit quick.

internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -60,193 +72,259 @@ func createDpuOperatorNameSpace() *corev1.Namespace {
...
+		LogLevel: 2,
+	}
+	return config
+}
+
+func createNameSpace(client client.Client, ns *v1.Namespace) {
+	// ignore error when createing the namespace since it can already exist
+	client.Create(context.Background(), ns)
+	found := v1.Namespace{}
+	err := client.Get(context.Background(), types.NamespacedName{Namespace: testNamespace, Name: ns.GetName()}, &found)
```
wizhaoredhat, line 90: We assume that namespaces are created instantly here. Would this always be the case? I expect similar code from `deleteNameSpace`

internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -9,18 +12,38 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	"k8s.io/client-go/tools/clientcmd"
+
+	configv1 "github.com/openshift/dpu-operator/api/v1"
+	//+kubebuilder:scaffold:imports
```
wizhaoredhat, line 46: Is this needed? What does it do for us.

internal/controller/dpuoperatorconfig_controller_test.go
bn222: It takes <1 sec.

internal/controller/dpuoperatorconfig_controller_test.go
bn222: Thanks for noticing. I've removed it.

internal/controller/dpuoperatorconfig_controller_test.go
bn222: Good point

PR #62, opened 2024-05-03
Allow users to set dev image tags from environment

SalDaniele:
It can be tedious modifying config/dev/local-images or having to apply local changes to build.

Allow users to just do:
DPU_OPERATOR_IMAGE="my-image" && make local-build


bn222:
Does this work?

PR #64, opened 2024-05-14
Add make fmt-check as a way to check if go fmt is needed

bn222:


Makefile
```diff
@@ -113,6 +113,13 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 fmt: ## Run go fmt against code.
 	go fmt ./...
 
+.PHONY: fmt-check
+fmt-check:
+	@files=$$(find . -name "*.go" -not -path "./vendor/*" -not -path "./dpu-api/vendor/*"); \
+	output=$$(gofmt -d $$files); \
```
wizhaoredhat, line 119: Typo? Shouldn't there be a space between go and fmt

bn222: no, gofmt is capable to do a dry run. `go fmt` isn't.

wizhaoredhat: Ah I see. you want to be able to dry run to see what go fmt will do to the code.

PR #77, opened 2024-05-29
Cleanup Device Plugin Code

wizhaoredhat:
- Remove use of fmt.Printf
- Simplify ListAndWatch

### DIFF
--- a/daemon/device-plugin/deviceplugin.go
+++ b/daemon/device-plugin/deviceplugin.go
@@ -27,7 +27,9 @@ const (
 	resourceName    = "openshift.io/dpu"
 )
 
-// sriovManager manages sriov networking devices
+type deviceList map[string]pluginapi.Device
+
+// nfResources manages NF networking devices
 type nfResources struct {
 	socketFile string
 	devices    map[string]pluginapi.Device // for Kubelet DP API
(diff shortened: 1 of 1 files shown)


daemon/device-plugin/deviceplugin.go
```diff
@@ -104,45 +115,40 @@ func (nf *nfResources) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequ
...
-
-func (nf *nfResources) Start() error {
-	nf.cleanup()
-	nf.ensureConnected()
-
-	ctx := context.Background()
+func (nf *nfResources) GetDevices() error {
+	err := nf.ensureConnected()
+	if err != nil {
+		return err
```
bn222, line 121: can we wrap the return here? `fmt.Errorf("... : %v", err)`

daemon/device-plugin/deviceplugin.go
```diff
@@ -104,45 +115,40 @@ func (nf *nfResources) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequ
...
 	}
 	nf.grpcServer = grpc.NewServer()
 
 	kubeletEndpoint := filepath.Join("unix:", DeprecatedSockDir, KubeEndPoint)
-
 	conn, err := grpc.Dial(kubeletEndpoint, grpc.WithTransportCredentials(insecure.NewCredentials()))
-
 	if err != nil {
-		fmt.Printf("%s device plugin unable connect to Kubelet : %v", resourceName, err)
+		nf.log.Error(err, "Device Plugin unable connect to Kubelet:", "resourceName", resourceName, "kubeletEndpoint", kubeletEndpoint)
```
bn222, line 151: Since we will return from `Start` here, we should not log the error here. We will log it at the `Start`-call site. Same remark for the rest of the PR.

daemon/device-plugin/deviceplugin.go
wizhaoredhat: Done

daemon/device-plugin/deviceplugin.go
wizhaoredhat: Done

daemon/device-plugin/deviceplugin.go
```diff
@@ -75,26 +75,40 @@ func (nf *nfResources) Changed() bool {
...
+	for {
+		if changed {
+			err := nf.sendDevices(stream)
+			if err != nil {
+				return err
+			}
+		}
+		time.Sleep(5 * time.Second)
+		changed = nf.changed()
+	}
```
bn222, line 89: Since this PR is already cleaning up the code, let's go one step further. The code in the change uses `nf.changed()` but also somehow manages changed state with a boolean (to for a change every time).  In addition, it has a bunch of side-effects saving the state to nf, but really that state is only needed to determine what to send to kube api.

I propose we rewrite all the other functions so that we can write something like this:
```
    var oldDevices []Device
    for {
        newDevices := getDevices()
        if !devicesEqual(oldDevices, newDevices) {
            if err := sendDevices(stream, newDevices); err != nil {
                return err
            }
            oldDevices = newDevices
        }
        time.Sleep(5 * time.Second)
    }
```

wizhaoredhat: Done

PR #78, opened 2024-06-07
Rework Device Plugin To Handle Multiple Device Types

wizhaoredhat:


daemon/device-plugin/deviceplugin.go
```diff
@@ -311,10 +265,11 @@ func (nf *nfResources) GetDevicePluginOptions(ctx context.Context, empty *plugin
...
-func NewGrpcPlugin() *nfResources {
+func NewDevicePlugin(dh DeviceHandler) *nfResources {
 	return &nfResources{
-		log:        ctrl.Log.WithName("DevicePlugin"),
-		devices:    make(map[string]pluginapi.Device),
-		socketFile: pluginEndpoint,
+		log:           ctrl.Log.WithName("DevicePlugin"),
+		devices:       make(map[string]pluginapi.Device),
+		socketFile:    pluginEndpoint,
+		deviceHandler: dh,
```
bn222, line 270: don't abbreviate

daemon/device-plugin/deviceplugin.go
```diff
@@ -311,10 +265,11 @@ func (nf *nfResources) GetDevicePluginOptions(ctx context.Context, empty *plugin
 	}, nil
 }
 
-func NewGrpcPlugin() *nfResources {
+func NewDevicePlugin(dh DeviceHandler) *nfResources {
```
bn222: Would it make sense to not get deviceHandler as an argument?

wizhaoredhat: Sorry what do you mean? We want to pass the device handler into the device plugin.

wizhaoredhat: We agreed to fix this along with main_test.go fixes.

PR #80, opened 2024-06-10
Fix DPU CNI binary copy regression

wizhaoredhat:
Due to: https://github.com/bn222/dpu-operator/blame/a1ff551f544c1705dfdb18cb67a093124640cfa9/daemon/main.go#L119
From commit: "35ebd23" Add SFC reconciler to daemon

### DIFF
--- a/daemon/main.go
+++ b/daemon/main.go
@@ -111,12 +111,12 @@ func main() {
 	err = prepareCni("/var/lib/cni/bin/dpu-cni")
 	if err != nil {
 		log.Error(err, "Failed to prepare CNI binary in /var/lib")
-		return
+		/* Don't return on error for now */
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
/cc @SalDaniele 
PTAL

bn222:
@wizhaoredhat I've purposefully added the return here. Can you elaborate what the issue is with the return?

wizhaoredhat:
> @wizhaoredhat I've purposefully added the return here. Can you elaborate what the issue is with the return?

On RHCOS, the directories exist in /opt but you can't write to it. If you tried to write, you will get a permission error. So for now we will just try to write and if it fails we log the error and move on.

PR #82, opened 2024-06-12
Add marvell comm channel support

bbhamidipati:
This commit adds a vendor gRPC server with Link Local (IPv6) support.


hw-vendor/marvell/bindata/daemonset.yaml
```diff
@@ -0,0 +1,68 @@
...
+        - name: host-run
+          mountPath: /var/run/netns
+          mountPropagation: Bidirectional
+        - name: proc
+          mountPath: /proc
+        # args:
+        # - --mode
+        # - {{.Mode}}
+      volumes:
+        - name: devicesock
```
wizhaoredhat, line 57: I don't think you need all these volume mounts.

hw-vendor/marvell/bindata/daemonset.yaml
```diff
@@ -0,0 +1,68 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: mrvl-comm-channel
+  # namespace: {{.Namespace}}
```
wizhaoredhat, line 6: The VSP daemonset should be rendered by the controller. Did Marvell have challenges working with the controller code?

hw-vendor/marvell/main.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+type MarvellConfig struct {
+	DeviceID string `yaml:"deviceID"`
+	VendorID string `yaml:"vendorID"`
+}
+
+func readConfig() MarvellConfig {
+	// read config file
+	config_path := "/config.yaml"
```
wizhaoredhat, line 30: Since this VSP is marvel specific. Why do we need to read from a file?

hw-vendor/marvell/README.md
```diff
@@ -0,0 +1,116 @@
...
+    ```
+
+## Launch Steps
+
+### Steps to Enable IPv6 Link Local
+
+1. **Edit Network Connection:**
+
+    ```sh
+    nmcli con edit type ethernet con-name ens81
```
wizhaoredhat, line 64: Shouldn't the VSP manage these connections? I.e. the VSP should be self contained and should require outside help from the controller/user for setting up the interfaces.

hw-vendor/marvell/README.md
```diff
@@ -0,0 +1,116 @@
...
+    ```sh
+    quit
+    ```
+
+To launch the Marvell Comm Channel:
+
+1. **Navigate to the Marvell Directory:**
+
+    ```sh
+    cd hw-vendor/marvell
```
wizhaoredhat, line 90: As mentioned prior, we should reconcile the daemonset in the controller.

hw-vendor/marvell/main.go
```diff
@@ -0,0 +1,264 @@
...
+			Port: ipPort.Port,
+		}, nil
+
+	}
+}
+
+// getDPUInterfaceName will return the VF1 Interface Name on DPU Side
+func getDPUInterfaceName() (string, error) {
+	// /sys/bus/pci/devices/0002:0f:00.1/net
+	baseNsenterCmd := "nsenter -t 1 -m -u -n -i -- "
```
wizhaoredhat, line 150: We can use the netlink library here instead of shelling. See SR-IOV CNI

hw-vendor/marvell/DockerFile.CommChannel.rhel
```diff
@@ -0,0 +1,58 @@
+# # Build the manager binary
```
wizhaoredhat, line 1: Please use the new path openshift-dpu-operator/internal/daemon/vendor-specific-plugins/marvell

hw-vendor/marvell/main.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	if err := scanner.Err(); err != nil {
+		return "", err
+	}
+	return IPv6Address, nil
+}
+
+func main() {
+	//UNIX socket setup
+	if err := os.RemoveAll(VendorPluginSocketPath); err != nil {
```
wizhaoredhat, line 246: Please look at openshift-dpu-operator/internal/daemon/vendor-specific-plugins/mock-vsp/mockvsp.go

internal/daemon/hostdaemon.go
```diff
@@ -127,7 +127,7 @@ func (d *HostDaemon) connectWithRetry() error {
 		  }
 		}]}`
 
-	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", d.addr, d.port), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithDefaultServiceConfig(retryPolicy))
+	conn, err := grpc.Dial(fmt.Sprintf("[%s]:%d", d.addr, d.port), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithDefaultServiceConfig(retryPolicy))
```
wizhaoredhat, line 130: Todo  @wizhaoredhat  Check if this works for IPv4.

internal/daemon/vendor-specific-plugins/marvell/DockerFile.CommChannel.rhel
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:22.04
+
+RUN DEBIAN_FRONTEND=noninteractive \
+  apt-get update \
+  && apt-get install net-tools kmod pciutils make iputils-ping iproute2 -y
```
wizhaoredhat, line 5: Todo Check if we can install this in a RHEL base image.

wizhaoredhat:
@thom311 Will be looking at this.

internal/daemon/vendor-specific-plugins/marvell/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+2. **Build the Project:**
+
+    ```sh
+    go build -o mrvl-vsp
+    ```
+
+## Docker Build Steps
+
+To create and push a Docker image:
```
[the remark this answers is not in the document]
hasan-alkama: These steps are needed on both side as vsp-daemon is common for host and DPU(arm64)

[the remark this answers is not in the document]
thom311: Right. In other words, I need to build amd64 and arm64 images, right?

I think the README should make that clearer (even if it seems obvious).

internal/daemon/vendor-specific-plugins/marvell/Dockerfile.mrvlVSP.rhel
```diff
@@ -0,0 +1,30 @@
+# Build Marvell VSP Binary
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.17 AS builder
```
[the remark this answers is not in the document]
bn222: This works if you're building it correctly

internal/daemon/vendor-specific-plugins/marvell/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Table of Contents
+
+- [Compilation Steps](#compilation-steps)
+- [Docker Build Steps](#docker-build-steps)
+- [Launch Steps](#launch-steps)
+
+## Compilation Steps
+
+Follow these steps to compile the Marvell VSP code:
```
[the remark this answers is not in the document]
bn222: Instead of writing code about steps how to build something in a readme, write those steps in a makefile and refer to it from the readme

internal/daemon/vendor-specific-plugins/marvell/README.md
```diff
@@ -0,0 +1,56 @@
...
+2. **Tag the Docker Image:**
+
+    ```sh
+    docker tag <image_id> <TagName:latest>
+    ```
+
+3. **Push the Docker Image:**
+
+    ```sh
+    docker push <TagName:latest>
```
bn222, line 46: Let's get rid of all these steps by moving them into the makefile

Dockerfile.mrvlVSP.rhel
```diff
@@ -0,0 +1,31 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.17 AS builder
```
bn222, line 1: 4.18

internal/daemon/device-handler/sriov-device-handler/sriovdevicehandler.go
```diff
@@ -123,10 +124,20 @@ func (s *sriovDeviceHandler) SetupDevices() error {
...
+	vendorName, _ := pi.Getvendorname()
+	var pfAddr string
+	if vendorName == "marvell" {
+		vendorID, deviceID, pfaddr, _ := pi.GetPcieDevFilter()
+		pfAddr = pfaddr
+		s.vfFilterFunc = CreatePcieDevFilter(vendorID, deviceID, "octeon_ep")
+	} else {
+		s.vfFilterFunc = CreatePcieDevFilter("8086", "145c", "idpf")
+		pfAddr = "0000:06:00.0"
+	}
```
bn222, line 138: @SalDaniele We need a hardware pool for Marvell in the ci.

wizhaoredhat:
Changes look good to me! @SalDaniele we should try the changes on the IPU to see if affects anything.

SalDaniele:
> Changes look good to me! @SalDaniele we should try the changes on the IPU to see if affects anything.

This is done by ci/prow/e2e-dpu-operator-tests, looks good

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -97,8 +100,12 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		logger.Error(err, "Failed to get DpuOperatorConfig resource")
 		return ctrl.Result{}, err
 	}
-
-	err := r.ensureDpuDeamonSet(ctx, dpuOperatorConfig)
+	err := r.ensureVSP(ctx, dpuOperatorConfig)
```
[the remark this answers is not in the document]
SalDaniele: Eventually, we do want to do something like this to deploy the vsp automatically from the operator (in fact, I have a PR submitted for this very thing at the moment: https://github.com/openshift/dpu-operator/pull/128 PTAL).

Currently, the user is expected to deploy the VSP prior to deploying the dpu operator / daemon. If this PR goes in, this will no longer work. I think the behavior should be to deploy the vsp if it is explicitly defined by the user, and otherwise, just deploy the daemon as before and expect that the user will handle deploying the vsp.

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -97,8 +100,12 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	if err != nil {
+		logger.Error(err, "Failed to ensure VSP is running")
+		return ctrl.Result{}, err
+	}
+	err = r.ensureDpuDeamonSet(ctx, dpuOperatorConfig)
```
SalDaniele, line 108: Please take a look at the failing test: https://github.com/openshift/dpu-operator/blob/main/internal/controller/dpuoperatorconfig_controller_test.go#L223

With this change the dpu daemon is not getting started properly, possibly due to an earlier failure in the controller when starting the vsp.

internal/platform/platform.go
```diff
@@ -25,11 +27,73 @@ func NewPlatformInfo() *PlatformInfo {
...
 		},
 	}
 }
 
+func (pi *PlatformInfo) Getvendorname() (string, error) {
+	klog.Infof("Detecting  Platform is DPU or not")
+	for _, detector := range pi.Detectors {
+		isDpu, err := detector.IsDpuPlatform()
+		if err != nil {
+			return "", err
```
SalDaniele, line 41: We need a way to detect the Vendor in the case we are running on the DPU as well.

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -184,6 +192,18 @@ func (r *DpuOperatorConfigReconciler) applyAllFromBinData(logger logr.Logger, bi
 	return nil
 }
 
+func (r *DpuOperatorConfigReconciler) ensureVSP(ctx context.Context, cfg *configv1.DpuOperatorConfig) error {
+	pi := platform.NewPlatformInfo()
```
SalDaniele, line 196: This code is currently called from the daemon, i.e. https://github.com/openshift/dpu-operator/blob/main/cmd/daemon/daemon.go#L57 which is running on the worker with the dpu installed.

This is being called here from the controller which is likely running on the control plane somewhere and not necessarily running on every dpu worker node. Are you sure the platform detection logic will work here?

SalDaniele (changes requested):
Overall, this looks good!

However, I do not think the logic to deploy the vsp from the operator is working. We cannot merge this with breaking changes.

One option might be to just remove deploying the vsp from this PR, and expect that the user will manually deploy the vsp prior to deploying the dpu operator as we currently do. When we have added the logic to the operator to also deploy the VSP, we can add an additional patch to ensure Marvell VSP deploys as expected. WDYT @bbhamidipati @bn222 @wizhaoredhat

Dockerfile.mrvlVSP.rhel
bbhamidipati: Currently TOT (Dockerfile.daemon.rhel) shows 4.17, can we move along with TOT?

internal/controller/dpuoperatorconfig_controller.go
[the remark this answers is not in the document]
bbhamidipati: > This will run on both the host side (OCP) and the dpu with the current design. I don't think this implementation is expected to work across both cases, is that correct?

VSP is expected on both.
On host side, it provides LifeCycleServices.
On DPU, it provides NF, BridgePort, etc.

> Eventually, we do want to do something like this to deploy the vsp automatically from the operator (in fact, I have a PR submitted for this very thing at the moment: https://github.com/openshift/dpu-operator/pull/128 PTAL).

This PR supports automatic deployment. We tested already.

> Currently, the user is expected to deploy the VSP prior to deploying the dpu operator / daemon. If this PR goes in, this will no longer work. I think the behavior should be to deploy the vsp if it is explicitly defined by the user, and otherwise, just deploy the daemon as before and expect that the user will handle deploying the vsp.

daemon depends on LifeCycleService. How it can work without VSP? Can you please elaborate?

internal/controller/dpuoperatorconfig_controller.go
bbhamidipati: I think platform is IPU.

This code expects vspDaemonImage = intel_VSP_IMAGE .
r.vspDaemonImage = os.Getenv(vendorName + "_" + "VSP_IMAGE")

In marvell case, it is like this:
config/dev/local-images-template.yaml:
+        - name: marvell_VSP_IMAGE
+          value: {{ .RegistryURL }}:5000/mrvl-vsp:dev

Is it the cause?

internal/platform/platform.go
bbhamidipati: Yes, its supported.

internal/controller/dpuoperatorconfig_controller.go
bbhamidipati: I suspect ensureVSP() is failing due to lack of env intel_VSP_IMAGE:
https://github.com/openshift/dpu-operator/pull/82/commits/ebbe5ee264b0f013531e41f05304c6de4b5d0e9c#diff-ac97373df236b7f27a14df2d71cf3b436098b60ff6ae8fccdb17c6a6266a33ecR204

if ensureVSP() fails, _err = r.ensureDpuDeamonSet(ctx, dpuOperatorConfig)_ won't be called.

We changed our code to log error instead of return error. 

This might help keep it running on IPU.

SalDaniele: Yes currently make-test does not test for the vsp deployment, this is still outside the scope of the dpu operator. It is just validating that the daemon is properly created when the operator config is added.

internal/controller/dpuoperatorconfig_controller.go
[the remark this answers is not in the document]
SalDaniele: Sorry, let me clarify.

Yes, VSP is expected on both. Currently we deploy the vsp manually, and then deploy the dpu-operator / daemon.

However, for the host side, the controller will not be running on every worker with a Marvell DPU installed, in fact it will likely be running on the control plane somewhere. The platform detection logic will not work in this case. As a result, we need to move the platform detection into the daemon, which will be running on the DPU workers with the proper PCI devices. PTAL at https://github.com/openshift/dpu-operator/pull/128, that is what I am trying to do here.

Dockerfile.mrvlVSP.rhel
wizhaoredhat: This is ok for now. We can adjust later.

PR #87, opened 2024-06-19
Move daemon into internal directory and to cmd/daemon/daemon.go

bn222:


api/v1/zz_generated.deepcopy.go
```diff
@@ -113,12 +113,27 @@ func (in *DpuOperatorConfigStatus) DeepCopy() *DpuOperatorConfigStatus {
 	return out
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
wizhaoredhat, line 116: Did you try building with these changes to deepcopy? It fails for me.

bn222: We need to fix the discrepancy between your env and mine in a separate PR before we tackle this new PR.

bn222: removing this change manually from the commit for now. We will bump to 1.22.3 early next week.

PR #89, opened 2024-06-20
Add SR-IOV numVF support

wizhaoredhat:


internal/daemon/device-handler/sriov-device-handler/sriovdevicehandler.go
```diff
@@ -75,11 +107,50 @@ func (nf *sriovDeviceHandler) GetDevices() (*dp.DeviceList, error) {
...
+		log: ctrl.Log.WithName("SriovDeviceHandler"),
 	}
+
+	// TODO: When changing the SRIOV numVfs, we should do the following:
+	// 1) Drain all pods running on the node with a drain controller running
+	// on the control plane. The nodes will be marked for draining and read by
+	// the drain controller.
+	// 2) Clearly define which PF and how many VFs from an API. User facing or
+	// otherwise
+	err := devHandler.SetupDevices()
```
bn222, line 150: Since we call SetupDevice here, it will call EnsureConnected. This introduces an order: daemon needs to be started after VSP, which we want to avoid.

Currently, you are printing an error but this means that the daemon goes into a broken state that we will need to recover by starting VSP and then restarting the daemon pod.

wizhaoredhat: Should we block for the VSP here? There is no integration with VSP yet, but I was wondering if this will be an issue if the VSP is part of the same pod as the daemon.

PR #90, opened 2024-06-21
Improve host daemon testing

bn222:


internal/daemon/hostdaemon.go
```diff
@@ -175,6 +176,7 @@ func (d *HostDaemon) cniCmdDelHandler(req *cnitypes.PodRequest) (*cni100.Result,
 }
 
 func (d *HostDaemon) Listen() (net.Listener, error) {
+	d.wg.Add(1)
```
wizhaoredhat, line 179: Shouldn't we call wg.Add right before we create the go routine?

bn222: Which go-routine?

PR #92, opened 2024-06-21
Fix race and introduce dpu daemon test

bn222:


internal/daemon/dpudaemon.go
```diff
@@ -125,96 +150,99 @@ func (d *DpuDaemon) Listen() (net.Listener, error) {
...
+	d.done <- nil
+	d.stopWg.Wait()
 }
 
-func (d *DpuDaemon) Stop() {
-	if d.server != nil {
-		d.server.GracefulStop()
-		d.server = nil
+func (d *DpuDaemon) getClient() *rest.Config {
+	if d.client == nil {
```
wizhaoredhat, line 226: This is weird an not obvious to an outside observer. Could we have a comment explaining this?

internal/daemon/path_manager.go
```diff
@@ -0,0 +1,33 @@
+package daemon
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/openshift/dpu-operator/dpu-cni/pkgs/cnitypes"
+)
+
+type PathManager struct {
```
wizhaoredhat, line 10: Why do we need this wrapper for Paths? When the paths can be made to consts.

internal/daemon/dpudaemon.go
bn222: I can move it to the constructor to simplify. Would that help?

internal/daemon/path_manager.go
bn222: This allows changing all the path by passing in a differnt `rootDir`. The paths used to be consts, but that means we can't change them dynamically.

internal/daemon/dpudaemon.go
```diff
@@ -125,96 +150,99 @@ func (d *DpuDaemon) Listen() (net.Listener, error) {
...
+
+		mgr, err := ctrl.NewManager(d.getClient(), ctrl.Options{
 			Scheme: scheme,
 			NewCache: func(config *rest.Config, opts cache.Options) (cache.Cache, error) {
 				opts.DefaultNamespaces = map[string]cache.Config{
 					"dpu-operator-system": {},
 				}
 				return cache.New(config, opts)
 			},
+			GracefulShutdownTimeout: &t,
```
wizhaoredhat, line 245: Could we have a comment explaining why we need this?

internal/daemon/dpudaemon.go
```diff
@@ -125,96 +150,99 @@ func (d *DpuDaemon) Listen() (net.Listener, error) {
...
-func (d *DpuDaemon) Serve(listen net.Listener) error {
-	d.server = grpc.NewServer()
-	pb.RegisterBridgePortServiceServer(d.server, d)
-	if err := d.server.Serve(listen); err != nil {
-		d.log.Error(err, "Failed to start serving")
-		return err
-	}
-	return nil
+func (d *DpuDaemon) Stop() {
+	d.done <- nil
```
wizhaoredhat, line 221: close(d.done)

internal/daemon/dpudaemon.go
```diff
@@ -125,96 +150,99 @@ func (d *DpuDaemon) Listen() (net.Listener, error) {
...
 		d.log.Error(err, "Failed to listen")
 		return err
 	}
 
-	wg.Add(1)
+	return d.Serve(listener)
+}
+
+func (d *DpuDaemon) Serve(listener net.Listener) error {
+	d.wg.Add(3)
```
wizhaoredhat, line 170: It is better convention to Add(1) next to each go func for better traceability. I think

internal/daemon/dpudaemon.go
bn222: Added comment to the code. Without this, the server cannot be stopped.

internal/daemon/dpudaemon.go
bn222: done

internal/daemon/dpudaemon.go
bn222: This would cause issues when the three go routines write to the channel as they stop. We still want to allow writing to the channel after `stop()` has been called, because that's part of the cleanup.

internal/daemon/dpudaemon.go
bn222: I've added this to `NewDpuDaemon`

PR #99, opened 2024-06-26
Introduce infrastructure needed to test against kubelet in kind

bn222:


dpu-cni/pkgs/cniserver/cniserver.go
```diff
@@ -71,16 +72,16 @@ func ensureRunDirExists(serverSocketPath string) error {
 
 // Listen creates a listener to a unix socket located in `socketPath`
 func (s *Server) Listen() (net.Listener, error) {
-	err := ensureRunDirExists(s.socketPath)
+	err := ensureRunDirExists(s.pathManager.CNIServerPath())
 	if err != nil {
 		return nil, fmt.Errorf("failed to create run directory for DPU CNI socket: %v", err)
 	}
-	listener, err := net.Listen("unix", filepath.Join(s.runDir, s.socketPath))
+	listener, err := net.Listen("unix", s.pathManager.CNIServerPath())
```
wizhaoredhat, line 79: This looks a bit weird since ensureRunDirExists was not changed in this PR but 
```
err := ensureRunDirExists(s.socketPath) != listener, err := net.Listen("unix", filepath.Join(s.runDir, s.socketPath))
```
should be different.
So I am not sure how s.pathManager.CNIServerPath() covers both when there is no input argument

wizhaoredhat: Unless runDir == ""

PR #103, opened 2024-06-28
Make DPU operator config cluster scope

bn222:


internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -115,7 +115,7 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 func (r *DpuOperatorConfigReconciler) createCommonData(cfg *configv1.DpuOperatorConfig) map[string]string {
 	// All the CRs will be in the same namespace as the operator config
 	data := map[string]string{
-		"Namespace":              cfg.Namespace,
```
wizhaoredhat, line 118: Shouldn't this still work?

api/v1/dpuoperatorconfig_types.go
```diff
@@ -41,6 +41,7 @@ type DpuOperatorConfigStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Cluster
```
wizhaoredhat, line 44: This is all that is needed to change the scope to "cluster scoped"?

bn222: yes

internal/controller/dpuoperatorconfig_controller.go
bn222: No, because the CR doesn't have a namespace.

api/v1/dpuoperatorconfig_types.go
wizhaoredhat: Shouldn't the RBAC needs to also change?

PR #105, opened 2024-06-28
Do not run test before images-push

bn222:


wizhaoredhat:
I do this all the time locally. So
/lgtm

PR #106, opened 2024-06-28
Autodetect mode on daemon

bn222:


internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -120,7 +120,7 @@ func (r *DpuOperatorConfigReconciler) createCommonData(cfg *configv1.DpuOperator
 	data := map[string]string{
 		"Namespace":              "openshift-dpu-operator",
 		"ImagePullPolicy":        r.imagePullPolicy,
-		"Mode":                   cfg.Spec.Mode,
+		"Mode":                   "auto",
```
wizhaoredhat, line 123: Is this intended to be hardcoded? Or left over from debug?

bn222: intented.

PR #108, opened 2024-07-01
Set seccompProfile in operator replicaset

SalDaniele:
We need to specify the seccompProfile. Without this, the replicaset fails to start via "make local-deploy"  with the following error:

Warning  FailedCreate  35s (x5 over 73s)  replicaset-controller  (combined from similar events): Error creating: pods "dpu-operator-controller-manager-5f95b6c647-bkf85" is forbidden: violates PodSecurity "restricted:latest": seccompProfile (pod or containers "manager", "kube-rbac-proxy" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")

### DIFF
--- a/bundle/manifests/dpu-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/dpu-operator.clusterserviceversion.yaml
@@ -38,7 +38,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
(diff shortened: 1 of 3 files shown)


SalDaniele:
@bn222 PTAL. Seems this is needed to start the dpu operator replicaset

PR #111, opened 2024-07-08
Add device plugin testing

wizhaoredhat:


internal/utils/path_manager.go
```diff
@@ -18,14 +22,12 @@ func (p *PathManager) CNIServerPath() string {
...
-	// SockDir is the default Kubelet device plugin socket directory
-	// SockDir = "/var/lib/kubelet/plugins_registry"
 	// The following path uses the deprecated Kubelet device plugin socket directory
-	return p.wrap("/var/lib/kubelet/device-plugins/sriovNet.sock")
+	return p.wrap("/var/lib/kubelet/device-plugins/kubelet.sock")
 }
 
 func (p *PathManager) PluginEndpoint() string {
-	return p.wrap("/var/lib/kubelet/device-plugins/sriovNet.sock")
+	return p.wrap(filepath.Join("/var/lib/kubelet/device-plugins", PluginEndpointSocket))
```
bn222, line 30: why not simply "/var/lib/kubelet/device-plugins/sriovNet.sock", like the rest?

internal/daemon/device-handler/mock-device-handler/mockdevicehandler.go
```diff
@@ -0,0 +1,37 @@
...
+	return nil
+}
+
+func (md *mockDeviceHandler) GetDevices() (*deviceplugin.DeviceList, error) {
+	devices := make(deviceplugin.DeviceList)
+
+	devices["ens5f0"] = pluginapi.Device{ID: "ens5f0", Health: pluginapi.Healthy}
+	devices["ens5f1"] = pluginapi.Device{ID: "ens5f1", Health: pluginapi.Healthy}
+	devices["ens5f2"] = pluginapi.Device{ID: "ens5f2", Health: pluginapi.Healthy}
+	devices["ens5f2"] = pluginapi.Device{ID: "ens5f3", Health: pluginapi.Healthy}
```
bn222, line 26:
```
devices := deviceplugin.DeviceList{
		"ens5f0": {ID: "ens5f0", Health: pluginapi.Healthy},
		"ens5f1": {ID: "ens5f1", Health: pluginapi.Healthy},
		"ens5f2": {ID: "ens5f2", Health: pluginapi.Healthy},
		"ens5f3": {ID: "ens5f3", Health: pluginapi.Healthy},
	}
```

internal/utils/path_manager.go
wizhaoredhat: Because we need the name of the socket file name elsewhere and not the path:
-		Endpoint:     dp.socketFile,
+		Endpoint:     utils.PluginEndpointSocket,

internal/daemon/device-handler/mock-device-handler/mockdevicehandler.go
wizhaoredhat: Done

internal/utils/path_manager.go
wizhaoredhat: This is just the API of the device plugin that the socket name is passed without the path.

bn222: I propose simplifying and making the code use the same style throughout (we have a similar requirement elsewhere). Elsewhere, we use the full path and then use `filepath` functions to derive a file name from the path. Otherwise, your API exposes the same (partial) information twice. The smaller the API surface, the less leaky it tends to be.

internal/daemon/device-plugin/deviceplugin.go
```diff
@@ -229,17 +230,17 @@ func (dp *dpServer) connectWithRetry(endpoint string) (*grpc.ClientConn, error)
...
-// 	}
+func (dp *dpServer) Stop() error {
+	dp.log.Info("Stopping Device Plugin...")
+	if dp.grpcServer == nil {
+		return nil
+	}
 
-// 	dp.grpcServer.Stop()
-// 	dp.grpcServer = nil
+	dp.grpcServer.Stop()
```
bn222, line 239: This isn't correct. We need to refactor and pull out `dp.grpcServer.Serve` from `RegisterDevicePlugin`, and properly implement stopping.

internal/utils/path_manager.go
wizhaoredhat: Done.

internal/daemon/device-plugin/deviceplugin.go
wizhaoredhat: I implemented the proper stopping. PTAL

internal/daemon/dpudaemon.go
```diff
@@ -207,6 +209,7 @@ func (d *DpuDaemon) Serve(listener net.Listener) error {
 
 	d.cancelManager = cancelManager
 
+	d.doneServing <- struct{}{}
```
bn222, line 218: This wont work. We need to use the other done channel since this doesn't handle an error in other goroutines.

wizhaoredhat: Done as discussed.

bn222:
I think we need to test against a testing vsp instead of a mock device handler.

wizhaoredhat:
> I favor testing against a test vsp instead of a mock device handler.

Done @bn222 , please look at the new mock VSP.

internal/daemon/hostdaemon.go
```diff
@@ -222,27 +217,48 @@ func (d *HostDaemon) ListenAndServe() error {
...
 		} else {
 			done <- nil
 		}
+		d.log.Info("Stopping manager")
 		wg.Done()
 	}()
 
-	err = <-done
+	// Run functions that handle registering of servers with services (e.g. kubelet)
+	err = d.dp.RegisterWithKubelet()
```
bn222, line 251: Why don't we do this inside the synchronous part of the dp?

internal/utils/path_manager.go
```diff
@@ -43,6 +47,46 @@ func (p *PathManager) CniPath(flavour Flavour) (string, error) {
...
 
+func (p *PathManager) VendorPluginSocket() string {
+	return p.wrap("/var/run/dpu-daemon/vendor-plugin/vendor-plugin.sock")
+}
+
 func (p *PathManager) wrap(path string) string {
 	return filepath.Join(p.rootDir, path)
 }
+
+// EnsureSocketDirExists makes sure that the socket being created is only accessible to root.
```
bn222, line 58: I like how you moved this to `PathManager`

wizhaoredhat: Thanks!

internal/daemon/hostdaemon.go
wizhaoredhat: We need to make sure we have started serving before registering with kubelet.

bn222: Agree, but this seems like something that is a detail that should be done inside the dp.

wizhaoredhat: Should we have some kind of Post Serve callback infrastructure?

bn222: Dicussed offline. Waiting for final changes

PR #113, opened 2024-07-11
Return an error if grpc connection to vsp fails

SalDaniele:
In the event we cannot establish connection with the vsp via grpc, we should report the error back up to improve debuggability.

### DIFF
--- a/internal/daemon/plugin/vendorplugin.go
+++ b/internal/daemon/plugin/vendorplugin.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"fmt"
 	"net"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


internal/daemon/plugin/vendorplugin.go
```diff
@@ -36,7 +36,11 @@ type GrpcPlugin struct {
 }
 
 func (g *GrpcPlugin) Start() (string, int32, error) {
-	g.ensureConnected()
+	err := g.ensureConnected()
+	if err != nil {
+		g.log.Error(err, "failed to ensure GRPC connection on grpcPlugin start")
```
wizhaoredhat, line 41: We should construct the error instead of logging a error message at each level. I think this is what we are trying to do with this repo. @bn222 Please correct me if we aren't on the same page on this.

bn222: Correct, the error should be passed up the stack

SalDaniele: Ok makes sense, updated to pass an Error back up

bn222: looks good to me, but want others to review too before merging

PR #114, opened 2024-07-12
Add detecting DPU

bn222:


wizhaoredhat:
I think we need an override for this that is optional as part of a CR. The reason is that it could be common for a customer to have a DPU that they are perhaps using for something else (i.e. SmartNIC/NIC mode) and only want one DPU to be used.

bn222:
This code only detects the devices, but doesn't do anything with it yet. We need to add code in the future to determine which devices we will configure/handle. Only when we have a full list of devices, can we decide which ones should not be handled. If we start handling configuration settings here, then the code will do two things: "detect and determine if we need to manage it or not".

cmd/daemon/daemon.go
```diff
@@ -55,7 +54,11 @@ func isDpuMode(log logr.Logger, mode string) (bool, error) {
 }
 
 func createDaemon(dpuMode bool, config *rest.Config) (Daemon, error) {
-	plugin := plugin.NewGrpcPlugin(dpuMode)
```
wizhaoredhat, line 58: The DPU Daemon tests now use NewGrpcPlugin. If this should be hidden away in the platform information (i.e. is not common for all DPUs) then we need to mock the platform class for testing.

internal/platform/platform.go
```diff
@@ -1,32 +1,82 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	var dpuDevices []ghw.PCIDevice
+	var detectors []VendorDetector
+	for _, pci := range pci.ListDevices() {
+		for _, detector := range pi.Detectors {
+			if detector.IsDPU(*pci) {
+				dpuDevices = append(dpuDevices, *pci)
+				detectors = append(detectors, detector)
```
wizhaoredhat, line 61: We should call `detectors` something like `activeDetectors` or `currentDetectors`.

internal/platform/platform.go
```diff
@@ -1,32 +1,82 @@
...
+		return plugin.NewGrpcPlugin(dpuMode), nil
+	} else {
+		dpuDevices, detectors, err := pi.listDpuDevices()
+		if err != nil {
+			return nil, errors.Errorf("Failed to get VspPlugin from platform: %v", err)
+		}
+		if len(dpuDevices) != 1 {
+			return nil, fmt.Errorf("Failed to detect which vsp to use")
+		}
+		return detectors[0].VspPlugin(dpuMode), nil
```
wizhaoredhat, line 80: We should make a TODO here. To support more than 1 VSP?

bn222: Do we want that? I think the error is what you will hit when you want to start adding more DPUs.

wizhaoredhat: Some text editors gives a list of TODOs... But I like to document things that appear weird to someone that doesn't have all the history. I.e. someone jumping into the project and not knowing the discussions our team has about what is mandatory and what is TODO in the future.

wizhaoredhat: It doesn't have to be called TODO. But just document oddities with the mindset if you knew nothing while reading the code. Then thinking it might be a bug.

bn222: What if we change the error to: `Multiple DPUs detected. Curreently only supporting 1 DPU per host`

internal/platform/platform.go
bn222: done

cmd/daemon/daemon.go
bn222: We definitely want to mock the platform instead of the grpc plugin.

internal/platform/platform.go
wizhaoredhat: Yes that would work too if you add a check for >1.

cmd/daemon/daemon.go
bn222: Currently, we don't test logic in cmd/daemon/daemon.go. I was planning to create a new struct that implements that logic. It's then when we will need to add tests for it.

SalDaniele:
FYI there is a bug introduced in this PR:

```
[root@localhost dpu-operator]# oc logs -n openshift-dpu-operator dpu-daemon-d45nh
2024-07-17T15:14:42Z    INFO    Daemon Init     Daemon init
2024-07-17T15:14:42Z    INFO    Daemon Init     Detected OpenShift      {"flavour": "MicroShift"}
2024-07-17T15:14:42Z    INFO    Daemon Init     Prepared CNI binary     {"path": "/opt/cni/bin/dpu-cni"}
2024-07-17T15:14:42Z    INFO    Daemon Init     Autodetected mode       {"isDPU": false}
2024-07-17T15:14:43Z    ERROR   Daemon Init     Failed to start daemon  {"error": "Failed to detect any DPU devices"}
main.main
        /workspace/cmd/daemon/daemon.go:165
runtime.main
        /usr/lib/golang/src/runtime/proc.go:271
```

PR #116, opened 2024-07-15
Return errors

bn222:
This also avoids printing a confusing error during testing when closing the server

### DIFF
--- a/dpu-cni/pkgs/sriov/sriov.go
+++ b/dpu-cni/pkgs/sriov/sriov.go
@@ -511,7 +511,7 @@ func (sm *sriovManager) CmdDel(req *cnitypes.PodRequest) error {
 		// Return nil when LoadConfFromCache fails since the rest
 		// of cmdDel() code relies on netconf as input argument
 		// and there is no meaning to continue.
-		klog.Errorf("Cannot load config file from cache %v", err)
+		klog.Warningf("Cannot load config file from cache: %v", err)
 		return nil
 	}
 
--- a/dpu-cni/pkgs/sriov/sriov.go
+++ b/dpu-cni/pkgs/sriov/sriov.go
@@ -504,23 +504,22 @@ func (sm *sriovManager) CmdDel(req *cnitypes.PodRequest) error {
 
 	netConf, cRefPath, err := sriovconfig.LoadConfFromCache(req.ContainerId, req.IfName)
 	if err != nil {
-		// If cmdDel() fails, cached netconf is cleaned up by
-		// the followed defer call. However, subsequence calls
-		// of cmdDel() from kubelet fail in a dead loop due to
-		// cached netconf doesn't exist.
-		// Return nil when LoadConfFromCache fails since the rest
-		// of cmdDel() code relies on netconf as input argument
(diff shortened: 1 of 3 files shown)


dpu-cni/example/cniserver_main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 		// Test NF CNI
 		err := cniserver.NewCNIServer(cniCmdNfAddHandler, cniCmdNfDelHandler).ListenAndServe()
 		if err != nil {
-			klog.Errorf("DPU CNI server Serve() failed: %v", err)
+			return fmt.Errorf("DPU CNI server Serve failed: %v", err)
```
SalDaniele, line 42: dpu-cni/example/cniserver_main.go:42:11: too many return values

bn222: ty done

dpu-cni/pkgs/sriov/sriov.go
```diff
@@ -511,8 +511,7 @@ func (sm *sriovManager) CmdDel(req *cnitypes.PodRequest) error {
 		// Return nil when LoadConfFromCache fails since the rest
 		// of cmdDel() code relies on netconf as input argument
 		// and there is no meaning to continue.
-		klog.Errorf("Cannot load config file from cache %v", err)
-		return nil
+		return fmt.Errorf("Cannot load config file from cache: %v", err)
```
wizhaoredhat, line 514: Hold. This cannot be changed. Please read the comment above.

dpu-cni/pkgs/sriov/sriov.go
```diff
@@ -504,23 +504,22 @@ func (sm *sriovManager) CmdDel(req *cnitypes.PodRequest) error {
...
 	netConf, cRefPath, err := sriovconfig.LoadConfFromCache(req.ContainerId, req.IfName)
 	if err != nil {
-		// If cmdDel() fails, cached netconf is cleaned up by
-		// the followed defer call. However, subsequence calls
-		// of cmdDel() from kubelet fail in a dead loop due to
-		// cached netconf doesn't exist.
-		// Return nil when LoadConfFromCache fails since the rest
-		// of cmdDel() code relies on netconf as input argument
-		// and there is no meaning to continue.
+		// This might happen if the cache has already been cleaned up. Kubelet
```
wizhaoredhat, line 507: Do we need to clean this up? I think we can leave it as is. It is easier to do a sync with upstream sriov-cni code base.

bn222: Good point. Do we want to copy things into this repo or do we want to make this a call into the other repo and vendor it in?

wizhaoredhat: Hmm tough call... Since I am looking at the SRIOV Env Variable issue. I can also look into this.

wizhaoredhat: Crap. I just remembered we had to hack the sriov-cni code because the Intel IPU doesn't support things. So vendoring approach won't work as is.

PR #117, opened 2024-07-16
Clean up local build targets

bn222:
This allows to build locally without having to manually change a file. This PR adds a tool that will write a kustomize config with the right arguments defined in the Makefile.

Also, clean up the targets and docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,8 +71,8 @@ CONTAINER_TOOL ?= podman
 
 # Use the image urls from the yaml that is used with Kustomize for local
 # development.
-DPU_OPERATOR_IMAGE := $(shell yq -r .spec.template.spec.containers[0].image config/dev/local-images.yaml)
-DPU_DAEMON_IMAGE := $(shell yq -r .spec.template.spec.containers[0].env[0].value config/dev/local-images.yaml)
+# DPU_OPERATOR_IMAGE := $(shell yq -r .spec.template.spec.containers[0].image config/dev/local-images.yaml)
+# DPU_DAEMON_IMAGE := $(shell yq -r .spec.template.spec.containers[0].env[0].value config/dev/local-images.yaml)
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -192,27 +186,48 @@ docker-push: ## Push docker image with the manager.
 	$(CONTAINER_TOOL) push ${IMG}
 
 GO_CONTAINER_CACHE = /tmp/dpu-operator-cache
+REGISTRY ?= $(shell hostname)
+# Use the image urls from the yaml that is used with Kustomize for local
+# development.
+DPU_OPERATOR_IMAGE := $(REGISTRY):5000/dpu-operator:dev
+DPU_DAEMON_IMAGE := $(REGISTRY):5000/dpu-daemon-image:dev
```
SalDaniele, line 193: I like the change, this will make my life a bit easier.

Maybe it would make sense to use ?= here to give users a knob to set DPU_OPERATOR_IMAGE / DPU_DAEMON_IMAGE in the event they are using quay or something instead of a local registry?

SalDaniele: Aahh disregard, I guess they can do this by adjusting the REGISTRY variable

Makefile
```diff
@@ -192,27 +186,48 @@ docker-push: ## Push docker image with the manager.
...
 	buildah manifest rm $(DPU_OPERATOR_IMAGE)-manifest || true
 	buildah manifest create $(DPU_OPERATOR_IMAGE)-manifest
 	buildah build --authfile /root/config.json --manifest $(DPU_OPERATOR_IMAGE)-manifest --platform linux/amd64,linux/arm64 -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
-	buildah manifest push --all $(DPU_OPERATOR_IMAGE)-manifest docker://$(DPU_OPERATOR_IMAGE)
 	buildah manifest rm $(DPU_DAEMON_IMAGE)-manifest || true
 	buildah manifest create $(DPU_DAEMON_IMAGE)-manifest
 	buildah build --authfile /root/config.json --manifest $(DPU_DAEMON_IMAGE)-manifest --platform linux/amd64,linux/arm64 -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.daemon.rhel -t $(DPU_DAEMON_IMAGE)
+
+.PHONY: local-pushx
+local-pushx: ## Push all container images necessary to run the whole operator
```
wizhaoredhat, line 225: Shouldn't we add "local-buildx" as a dependency to "local-pushx"

bn222: done

PR #119, opened 2024-07-17
Fix bug in DPU Platform Info

SalDaniele:
We need to create a new instance of the PlatformInfo struct to ensure the NewIntelDetector is added.

Additionally, fix a small bug in determining a device is an Intel DPU. The Intel IPU pci Class/Vendor/Product should be:

Network controller Intel Corporation Infrastructure Data Path Function

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -49,9 +49,6 @@ func (pi *PlatformInfo) listDpuDevices() ([]ghw.PCIDevice, []VendorDetector, err
 		return nil, nil, errors.Errorf("Error getting PCI info: %v", err)
 	}
 
-	if err != nil {
-		return nil, nil, err
-	}
 	var dpuDevices []ghw.PCIDevice
 	var activeDetectors []VendorDetector
 	for _, pci := range pci.ListDevices() {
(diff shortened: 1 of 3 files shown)


SalDaniele:
@bn222 PTAL

internal/platform/platform.go
```diff
@@ -74,7 +88,25 @@ func (pi *PlatformInfo) VspPlugin(dpuMode bool) (*plugin.GrpcPlugin, error) {
...
 			return nil, fmt.Errorf("Failed to detect any DPU devices")
 		}
-		// FIXME: We need to find a good way to determine VF from PF with the Intel DPU versus different DPU vendors.
-		return detectors[0].VspPlugin(dpuMode), nil
+
+		var pfDevices []*ghw.PCIDevice
+		var pfDetectors []VendorDetector
+
+		for i, device := range dpuDevices {
+			isVF, err := isVirtualFunction(device.Address)
```
wizhaoredhat, line 96: We might need this logic in listDpuDevices instead...

internal/platform/platform.go
```diff
@@ -63,6 +65,18 @@ func (pi *PlatformInfo) listDpuDevices() ([]ghw.PCIDevice, []VendorDetector, err
 	return dpuDevices, activeDetectors, nil
 }
 
+func isVirtualFunction(device string) (bool, error) {
```
wizhaoredhat, line 68: I think this belongs to ipu.go.

wizhaoredhat:
since Sal is on PTO. Let's merge this in and fix the issues above in a seperate PR.

PR #122, opened 2024-07-26
IIC-247: Deploy vsp from dpu operator

SalDaniele:


SalDaniele:
@wizhaoredhat PTAL.

Still a WIP, need to add tetss to verify the vsp is coming up, but tested and this works / solves the issues with component ordering.

Hoping to get some initial feedback on whether it makes sense to deploy the vsp at this point.

One problem I am trying to figure out is that we will need to know the specific vendor plugin to deploy prior to running the platform detection in the daemon...I don't believe the controller has this information at the time of deploying the vsp.

SalDaniele:
Closing, will re-open a PR with a modified implementation

PR #124, opened 2024-08-06
proto file change for SetNumVfs

sudhar-krishnakumar:
proto file change for SetNumVfs


wizhaoredhat:
@bn222
I had some chats with Sudhar for this change. I think having an API that the DPU Operator can call for the VSP to handle the SR-IOV interfaces is the proper approach.

SetNumVfs will set the requested number of VFs. The API will return back, the actual number of VFs being configured. This will allow the user facing CR to set the VF to any number without worrying about any webhook validation.

PR #125, opened 2024-08-12
fix image references in 4.17

ashwindasr:
Fix incorrect image references in `release-4.17`

Please verify and follow up if other places need to be modified as well.


ashwindasr:
Raised in correctly to main, since TP automation does not forward commits to 4.17 now

https://github.com/openshift/dpu-operator/pull/127

PR #126, opened 2024-08-12
NO-JIRA: fix image references in 4.18

ashwindasr:
Fix incorrect image references in `release-4.18`

Please verify and follow up if other places need to be modified as well.


ashwindasr:
I assuming TP automation now forwards to 4.18+

PR #134, opened 2024-08-26
Add IPU-specific VSP build support

SamD2021:
This functionality is currently either being done manually or through other automation tools. By bringing Vendor Specific Plugins closer to the DPU, we can make it is easier to test, develop and deploy.
This implimentation is currently IPU-specific as its the only one we have currently, but can be expanded to house other VSPs. 
Starting the vsp plugin should still be handled elsewhere for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,6 +247,32 @@ docker-buildx: test ## Build and push docker image for the manager for cross-pla
 	- $(CONTAINER_TOOL) buildx rm project-v3-builder
 	rm Dockerfile.cross
 
+# Build and Push VSP
+# NOTE: This implementation is IPU-specific since its the only vendor specific plugin at this time, but can expand when others are available.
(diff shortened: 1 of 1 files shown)


SamD2021:
@wizhaoredhat Please let me know if there is anything extra that you think this implementation should do!

wizhaoredhat:
@SamD2021 , can't we have a rule that scans through the directory and builds the VSP regardless of vendor?

SamD2021:
> @SamD2021 , can't we have a rule that scans through the directory and builds the VSP regardless of vendor?

We could, I'm just not sure what files it should scan for since Sal said it doesn't seem to have a standard file structure or naming scheme for the plugins yet. Is there a specific structure they all follow that i could search for yet?

wizhaoredhat:
Please take a look at https://github.com/openshift/dpu-operator/pull/82

SalDaniele:
Closing for now, will need Intel's help adding this similar to #82

PR #135, opened 2024-08-28
move venv, pull_secret out of tree, and update cda

bn222:


vrindle:
make test runs on 140 which is down

PR #138, opened 2024-08-29
Add env substitution for path

vrindle:


bn222:
Do we need `envsubst` here?

bn222:
I think that something simple like https://github.com/openshift/dpu-operator/pull/139 should be sufficient

PR #142, opened 2024-09-02
Add deploy traffic flow test target to Makefile

vrindle:


hack/ipu_deploy.sh
```diff
@@ -2,6 +2,8 @@
 cd cluster-deployment-automation
 source ocp-venv/bin/activate
 
+cp /root/pull_secret.json .
+
```
bn222, line 6: not needed

cluster_configs/config.yaml
```diff
@@ -0,0 +1,25 @@
...
+    duration: "30"
+    # Location of artifacts from run can be specified: default <working-dir>/ft-logs/
+    # logs: "/tmp/ft-logs"
+    connections:
+      - name: "Connection_1"
+        # supported types: iperf-tcp, iperf-udp, netperf-tcp-stream, netperf-tcp-rr
+        type: "iperf-tcp"
+        instances: 1
+        server:
+          - name: "$worker"
```
bn222, line 15: @wizhaoredhat can you PTAL at this line?

bn222:
@vrindle rebase please. Also remove `envsubst`

cluster_configs/config.yaml
bn222: lgtm for now, but we want tft to be able to have something like "{{worker(0)}}" in place that will query the cluster with `oc`

PR #148, opened 2024-09-05
CDA: hardcode current iso

SalDaniele:
Previously testing would call out to CDA to determine the iso to install with.

Instead, define this in dpu operator so we can track updates / validate when we move to newer versions.

TODO: Instead of using a "pre-built" iso, we should eventually strive to automatically create the iso from the latest nightly.


SalDaniele:
Closing, dup of #146

PR #162, opened 2024-09-23
Add GetDevice() to Marvell VSP

alkamah-marvell:
This commit adds GetDevice Support for mrvl-vsp

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -33,19 +34,172 @@ const (
...
+	nfInterfaceName := fmt.Sprintf("nf_interface%d", index)
+	dpInterfaceName := fmt.Sprintf("dp_interface%d", index)
+	vethLink := &netlink.Veth{
+		LinkAttrs: netlink.LinkAttrs{Name: nfInterfaceName},
+		PeerName:  dpInterfaceName,
+	}
+	if err := netlink.LinkAdd(vethLink); err != nil {
+		return err
+	}
+	if err := netlink.LinkSetUp(vethLink); err != nil {
```
thom311, line 95: Below we call `LinkSetUp()` on the `peerLink` instance that we just looked up fresh with `LinkByName()`.

I think that should also be done in this case. I mean, move this before `netlink.LinkSetUp(peerLink)` and operate on `nfLink`.

Otherwise, `LinkSetUp(vethLink)` needs internally call `LinkByName()` to find the ifindex. Instead, by using `LinkSetUp(nfLink)`, the ifindex is already known.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -33,19 +34,172 @@ const (
...
+	if vsp.deviceStore != nil {
+		for nfMacAddress, mrvlDeviceInfo := range vsp.deviceStore {
+			nfLink, err := netlink.LinkByName(mrvlDeviceInfo.nfInterfaceName)
+			if err != nil {
+				klog.Errorf("Error occurred in getting Link By Name: %v", err)
+				return err
+			}
+			if err := netlink.LinkDel(nfLink); err != nil {
+				klog.Errorf("Error occurred in deleting Link: %v", err)
+				return err
```
thom311, line 139: Shouldn't the cleanup proceed trying to delete all interfaces, even if errors happen? Aborting on the first error seems wrong, especially since iterating over `range vs.deviceStore` iterates in an arbitrary order.

Also, I think the map must not be modified (`deletet()`) while iterating over it.

I think the errors should be all collected, but the code should also try to delete all interfaces:
```go
    var errResult error
    if vsp.deviceStore != nil {
        deviceStore := vsp.deviceStore
        vsp.deviceStore = nil
        for nfMacAddress, mrvlDeviceInfo := range deviceStore {
             nfLink, err := netlink.LinkByName(mrvlDeviceInfo.nfInterfaceName)
             if err != nil {
                 klog.Errorf("Error occurred in deleting Link: %v", err)
                 errResult = errors.Join(errResult, err)
                 continue
             }
             ...
         }
    }
    return errResult
```

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -350,27 +526,39 @@ func (vsp *mrvlVspServer) Serve(listener net.Listener) error {
...
+func WithPathManager(pathManager utils.PathManager) func(*mrvlVspServer) {
+	return func(vsp *mrvlVspServer) {
+		vsp.pathManager = pathManager
+	}
+}
+
 func NewMarvellVspServer(opts ...func(*mrvlVspServer)) *mrvlVspServer {
-	vsp := &mrvlVspServer{
-		log: ctrl.Log.WithName("MarvellVsp"),
+	mrvlVspServerInstance = &mrvlVspServer{
```
thom311, line 541: this `New*()` method stores a pointer to the instance in a global variable. This seems a bad name for what it does.

Regardless, I don't understand why this global exists at all. I think it should be dropped, and places that require this instance, should get it passed down as an argument by the caller. The only user of the global is `Listen()`, which already has this instance via the `(vsp *mrvlVspServer)` parameter.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -33,19 +34,172 @@ const (
...
+	err = yaml.Unmarshal(data, &config)
+	if err != nil {
+		return deviceConfig{}, err
+	}
+	return config, nil
+}
+
+// createVethPair function to create a veth pair with the given index and InterfaceInfo
+func (vsp *mrvlVspServer) createVethPair(index int) error {
+	nfInterfaceName := fmt.Sprintf("nf_interface%d", index)
```
wizhaoredhat, line 86: Can you explain what interfaces we are creating here in commentary. It is not clear what this function is doing although it implies creating a Veth pair.
E.g. What is nf_interface, dp_interface, nfInterfaceName, dpInterfaceName, etc...

internal/daemon/vendor-specific-plugins/marvell/main.go
wizhaoredhat: I agree that errors should be collected but this function should proceed for interfaces.

internal/daemon/vendor-specific-plugins/marvell/config.yaml
```diff
@@ -0,0 +1,2 @@
+portType: "veth"  # Currently supported values are "veth","hwlbk"
```
wizhaoredhat, line 1: Can we hardcode these values for now? (instead of having a config file, or is this necessary for debugging?)

The proper way is to create a CR or config map from the Operator side. But we need a proper proposal of what is needed, I don't think we are ready for this yet.

internal/daemon/vendor-specific-plugins/marvell/main.go
wizhaoredhat: Agreed we don't really need to store this in a global. We don't do this elsewhere in the code.

internal/daemon/vendor-specific-plugins/marvell/main.go
alkamah-marvell: Okay so if i do LinkSetUp() on nflink after LinkByName() would be sufficient No need for vethLink 
is that what you mean?

thom311: `vethLink` is still needed for `LinkAdd()`.

I mean
```diff
From ebe1cff131ac5e451f8529cd61b18de10fc48c42 Mon Sep 17 00:00:00 2001
From: Thomas Haller <thaller@redhat.com>
Date: Tue, 1 Oct 2024 11:26:48 +0200
Subject: [PATCH 1/1] fixup! Add GetDevice() to Marvell VSP This commit adds
 GetDevice Support for mrvl-vsp

---
 internal/daemon/vendor-specific-plugins/marvell/main.go | 7 ++++---
 1 file changed, 4 insertions(+), 3 deletions(-)

diff --git a/internal/daemon/vendor-specific-plugins/marvell/main.go b/internal/daemon/vendor-specific-plugins/marvell/main.go
index 51ba3c7ab12b..d7c1c7ab9659 100644
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -85,34 +85,35 @@ func readConfig() (deviceConfig, error) {
 func (vsp *mrvlVspServer) createVethPair(index int) error {
 	nfInterfaceName := fmt.Sprintf("nf_interface%d", index)
 	dpInterfaceName := fmt.Sprintf("dp_interface%d", index)
 	vethLink := &netlink.Veth{
 		LinkAttrs: netlink.LinkAttrs{Name: nfInterfaceName},
 		PeerName:  dpInterfaceName,
 	}
 	if err := netlink.LinkAdd(vethLink); err != nil {
 		return err
 	}
-	if err := netlink.LinkSetUp(vethLink); err != nil {
-		return err
-	}
 
 	nfLink, err := netlink.LinkByName(nfInterfaceName)
 	if err != nil {
 		return err
 	}
 
 	peerLink, err := netlink.LinkByName(dpInterfaceName)
 	if err != nil {
 		return err
 	}
 
+	if err := netlink.LinkSetUp(nfLink); err != nil {
+		return err
+	}
+
 	if err := netlink.LinkSetUp(peerLink); err != nil {
 		return err
 	}
 
 	vsp.deviceStore[nfLink.Attrs().HardwareAddr.String()] = mrvlDeviceInfo{
 		nfInterfaceName: nfInterfaceName,
 		dpInterfaceName: dpInterfaceName,
 		dpMAC:           peerLink.Attrs().HardwareAddr.String(),
 		health:          "Healthy",
 	}
-- 
2.46.2

```

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -312,13 +468,12 @@ func (vsp *mrvlVspServer) Listen() (net.Listener, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to listen on the vendor plugin socket: %v", err)
 	}
-
+	mrvlVspServerInstance := &mrvlVspServer{}
```
thom311, line 471: I think, it should not create a new instance here. Instead, the caller `main()` already has a `mrvlVspServer` variable. Let it pass it to `Listen()` as argument.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -363,14 +531,12 @@ func NewMarvellVspServer(opts ...func(*mrvlVspServer)) *mrvlVspServer {
 }
 
 func main() {
-	// mrvlVspServer := commchannel.NewMarvellVspServer()
 	mrvlVspServer := NewMarvellVspServer()
-	listener, err := mrvlVspServer.Listen()
+	listener, err := mrvlVspServer.Listen(mrvlVspServer)
```
thom311, line 535: Minor complained:

I know I suggested this, but the `mrvlVspServer` argument is unnecessary here.

note that `Listen()` already has:
```
func (vsp *mrvlVspServer) Listen(mrvlVspServerInstance *mrvlVspServer) (net.Listener, error)
```

We don't need the variable `mrvlVspServer`. We already have the `vsp` argument, the *self* pointer for the method call.

wizhaoredhat: Talked about during the meeting that the input arg should be removed.

thom311 (approved):
lgtm

wizhaoredhat:
also LGTM. @bn222 / @SalDaniele PTAL.

PR #163, opened 2024-09-23
[th/vendor-plugin-deploy] various fixes for starting the Marvell VSP on the DPU

thom311:
With #128 merged, there is now a way for starting the VSP plugin from the operator.

However, there are various missing pieces.

Fix some of the issues, in particular to fix starting the Marvell VSP in the DPU.

CC: @bbhamidipati , @alkamah-marvell , @hasan-alkama (FYI)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -96,15 +96,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	vspImages := make(map[string]string)
-
-	for _, vspImageName := range plugin.VspImages {
-		value := os.Getenv(vspImageName)
-		if value == "" {
-			setupLog.Info("VspImage env var not set", "VspImage", vspImageName)
-		}
-		vspImages[vspImageName] = value
-	}
+	vspImages := plugin.CreateVspImagesMap(true, setupLog)
 
 	dpuDaemonImage := os.Getenv("DPU_DAEMON_IMAGE")
 	if dpuDaemonImage == "" {
--- a/config/dev/local-images-template.yaml
+++ b/config/dev/local-images-template.yaml
@@ -22,6 +22,8 @@ spec:
           value: {{ .RegistryURL }}:5000/dpu-daemon:dev
         - name: IntelVspImage
           value: {{ .RegistryURL }}:5000/intel_vsp:dev
+        - name: MarvellVspImage
+          value: {{ .RegistryURL }}:5000/mrvl-vsp:dev
         - name: IMAGE_PULL_POLICIES
           value: Always
         image: {{ .RegistryURL }}:5000/dpu-operator:dev
--- a/internal/platform/ipu.go
+++ b/internal/platform/ipu.go
@@ -58,8 +58,8 @@ func (pi *IntelDetector) IsDpuPlatform() (bool, error) {
 }
 
 func (pi *IntelDetector) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client) *plugin.GrpcPlugin {
-	intelVspImage := vspImages["IntelVspImage"]
-	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVspImage(intelVspImage))
+	template_vars := plugin.CreateVspImageVars(vspImages["IntelVspImage"])
+	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVspImage(template_vars))
 }
 
 func (d *IntelDetector) GetVendorName() string {
--- a/internal/platform/ipu.go
+++ b/internal/platform/ipu.go
@@ -59,6 +59,8 @@ func (pi *IntelDetector) IsDpuPlatform() (bool, error) {
 
 func (pi *IntelDetector) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client) *plugin.GrpcPlugin {
 	template_vars := plugin.CreateVspImageVars(vspImages["IntelVspImage"])
+	template_vars["Command"] = `[ "/usr/bin/ipuplugin" ]`
+	template_vars["Args"] = `[ "-v=debug" ]`
 	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVspImage(template_vars))
 }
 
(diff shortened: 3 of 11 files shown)


internal/platform/ipu.go
```diff
@@ -58,8 +58,11 @@ func (pi *IntelDetector) IsDpuPlatform() (bool, error) {
 }
 
 func (pi *IntelDetector) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client) *plugin.GrpcPlugin {
-	intelVspImage := vspImages["IntelVspImage"]
-	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVspImage(intelVspImage))
+	image := vspImages["IntelDetector"]
+	render_data := plugin.CreateCommonData(image)
+	render_data["Command"] = `[ "/usr/bin/ipuplugin" ]`
+	render_data["Args"] = `[ "--bridgeType=linux", "--interface=enp0s1f0d3", "--portMuxVsi=0x0e", "--ovsCliDir=/opt/p4/p4-cp-nws", "-v=debug" ]`
```
SalDaniele, line 64: According to Intel, these args are no longer required....we can leave debug just for debug-ability: https://github.com/intel/ipu-opi-plugins/commit/5d07f00942448bd02dbd54dd0207696a537f64e6#diff-4de381419f5248baa3c90d5e7189bb5770d291fa40f7ba3c0918c31b6f42d61dR32

internal/platform/marvell-dpu.go
```diff
@@ -52,8 +52,11 @@ func (pi *MarvellDetector) IsDpuPlatform() (bool, error) {
 }
 
 func (pi *MarvellDetector) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client) *plugin.GrpcPlugin {
-	MarvellVspImage := vspImages["MarvellVspImage"]
-	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVspImage(MarvellVspImage))
+	image := vspImages["MarvellVspImage"]
+	render_data := plugin.CreateCommonData(image)
+	render_data["Command"] = `[ "/vsp-mrvl" ]`
+	render_data["Args"] = `[ "--bridgeType=linux", "--interface=enp0s1f0d3", "--portMuxVsi=0x0e", "--ovsCliDir=/opt/p4/p4-cp-nws", "-v=debug" ]`
```
SalDaniele, line 58: These are (were) Intel specific arguments. Looking at Marvell's PR, these are not required: https://github.com/openshift/dpu-operator/commit/1b4275bb1b35367b0e9805aa9a6be0ad4cbfbea0#diff-f83f7283821d3bddf407b05738df9d57313ebc1ef62525d7a6ddabe16e192c3dR1

config/dev/local-images-template.yaml
```diff
@@ -22,6 +22,8 @@ spec:
           value: {{ .RegistryURL }}:5000/dpu-daemon:dev
         - name: IntelVspImage
           value: {{ .RegistryURL }}:5000/intel_vsp:dev
+        - name: MarvellVspImage
+          value: {{ .RegistryURL }}:5000/mrvl-vsp:dev
```
SalDaniele, line 26: If we want this to build and deploy automatically with our Makefile, this will need to match the env here: https://github.com/openshift/dpu-operator/blob/main/Makefile#L242

internal/platform/platform.go
```diff
@@ -132,24 +148,32 @@ func (pi *PlatformInfo) listDpuDevices() ([]ghw.PCIDevice, []VendorDetector, err
...
+}
+
 func (pi *PlatformInfo) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client) (*plugin.GrpcPlugin, error) {
 	var detector VendorDetector
+	var err error
+
 	if dpuMode {
-		// TODO: We need to also detect the platform in dpuMode
-		detector = NewIntelDetector()
+		detector, err = pi.detectDpuPlatform()
```
SalDaniele, line 171: I don't think this code path has been exercised before for the IPU, we should make sure this is green in the e2e CI run.

SalDaniele (commented):
This LGTM! We should validate these changes on Marvell hardware before merging, we need to add this to our e2e tests / lab.

Also we should include a bump of CDA in this patchset where we remove building and deploying the Marvell VSP from the post config steps.

SalDaniele:
@bbhamidipati PTAL

config/dev/local-images-template.yaml
thom311: I don't follow. Could you elaborate?

internal/platform/platform.go
thom311: does the github action on this PR contains such a test?

config/dev/local-images-template.yaml
SalDaniele: Sorry disregard, I was looking at the wrong thing!

internal/platform/platform.go
SalDaniele: Yes, this will be tested by ci/prow/e2e-dpu-operator-tests. I don't see any reason this wouldn't work, but since we have hardcoded this in the past we should validate this is green on both Intel and Marvell hardware

cmd/main.go
```diff
@@ -96,15 +96,7 @@ func main() {
...
-	vspImages := make(map[string]string)
-
-	for _, vspImageName := range plugin.VspImages {
-		value := os.Getenv(vspImageName)
-		if value == "" {
-			setupLog.Info("VspImage env var not set", "VspImage", vspImageName)
-		}
-		vspImages[vspImageName] = value
-	}
+	vspImages := plugin.CreateVspImagesMap(true, setupLog)
```
SalDaniele, line 99: Thanks for fixing this

internal/platform/ipu.go
```diff
@@ -58,8 +58,10 @@ func (pi *IntelDetector) IsDpuPlatform() (bool, error) {
 }
 
 func (pi *IntelDetector) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client) *plugin.GrpcPlugin {
-	intelVspImage := vspImages["IntelVspImage"]
-	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVspImage(intelVspImage))
+	template_vars := plugin.CreateVspImageVars(vspImages["IntelDetector"])
```
SalDaniele, line 61: Shouldn't this be "IntelVspImage"?

thom311: of course... thanks.

(reminds me to when I suggested not to use string literals, because they are error prone).

PR #160, opened 2024-09-19
Fix typo calling tft-test

vrindle:

PR #164, opened 2024-09-24
[th/disable-tft-tests] Makefile: disable "deploy_tft_tests" step from "e2e-test" for now

thom311:
It is currently not working yet. Disable it for now, otherwise the "e2e-test" target will fail.

Note that currently there is a bug in "deploy_tft_tests" target, so currently it is also not called. So this change has little effect. However, it will have effect, when that bug is fixed (and the target starts to fail for real).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ ipu_deploy:
 	bash $(DEPLOY_TFT_SCRIPT)
 
 .PHONY: e2e_test
-e2e-test: ipu_host ipu_deploy deploy_tft_tests
+e2e-test: ipu_host ipu_deploy
 	@echo "E2E Test Completed"
 
 .PHONY: all


bn222:
Can we re-enable this after the bug has been fixed? I would rather have it fail since we need to know that too.

thom311:
> Can we re-enable this after the bug has been fixed?

Yes. The commit message and the PR text suggests to do that (with text like "disable it for NOW", suggesting to re-enable it later).

thom311:
> Can we re-enable this after the bug has been fixed?

hm, or... which bug to you mean? The one from https://github.com/openshift/dpu-operator/pull/160 ?

No, fixing *that* bug should not mean to re-enable the test. 

That test was added and it never worked. It should not have been merged in such state, and it should not be enabled until it starts to properly work. Once *that* bug is fixed, it should be enabled.

PR #166, opened 2024-09-25
[th/hack-scripts] cleanup hack scripts

thom311:
- hack: make scripts executable and add shebang
- hack: use "set -e" to fail scripts on error
- Makefile: directly call scripts instead of using variables
- hack: use local temporary directory instead of shared one in /tmp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ doc/*.png
 *.swp
 *.swo
 *~
+
+/.tmp/
--- a/hack/deploy_traffic_flow_tests.sh
+++ b/hack/deploy_traffic_flow_tests.sh
@@ -2,8 +2,9 @@
 
 set -e
 
+source ./.tmp/ocp-venv/bin/activate
+
 cd ocp-traffic-flow-tests
-source /tmp/ocp-venv/bin/activate
 
 export KUBECONFIG=/root/kubeconfig.ocpcluster
 nodes=$(oc get nodes)
--- a/hack/deploy_traffic_flow_tests.sh
+++ b/hack/deploy_traffic_flow_tests.sh
@@ -1,5 +1,5 @@
 cd ocp-traffic-flow-tests
-source ocp-venv/bin/activate
+source /tmp/ocp-venv/bin/activate
 
 export KUBECONFIG=/root/kubeconfig.ocpcluster
 nodes=$(oc get nodes)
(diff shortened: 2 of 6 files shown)


hack/deploy_traffic_flow_tests.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
bn222, line 1:
```
/usr/bin/env bash
```

.gitignore
```diff
@@ -25,3 +25,5 @@ doc/*.png
 *.swp
 *.swo
 *~
+
+/.tmp/
```
bn222, line 29: Let's not use an in-tree directory.

thom311: The change is done for reasons explained in the commit message. Primarily, so that you can clone the git project at multiple locations, and running make targets does not interfere with each other. Whether that ultimately will work is unclear, there might be other obstacles. But sharing one directory in `/tmp/` is a definite problem in this regard.

How do you suggest to do that, without using a temporary directory that is clearly tied to the working directory of the current checkout? Do you just want to reject that use case?

Also, what is the problem with the in-tree directory? Python tools create them plenty (`.mypy_cache`, `__pycache__/`), various of our setup scripts also create them (`ocp-venv`) ([example1](https://github.com/bn222/cluster-deployment-automation/blob/16dd5a6cc83c3558ce008b6a0bd432900af76646/README.md#install-software-and-python-packages-start-a-python-virtual-environment-and-setup), [example2](https://gitlab.cee.redhat.com/ccit/jenkins-csb-customers/nst-casc-net-hw-ci/-/blob/fb620ce1beabf3ef6b4c994040972e193a9ad1a2/scripts/ovn-traffic-flow/traffic-flow-run-test.jenkinsfile#L28)).

It also is a benefit that you see the temporary data with `git clean -ndx` and can wipe with with `git clean -fdx`, without having state left at unspecified places.

hack/deploy_traffic_flow_tests.sh
thom311: I choose `#!/bin/bash` intentionally and I am away of `env` (and decided intentionally against it for various reasons).

If you suggest an alternative, then it would be useful to leave a hint for why that is preferable.

OK, changed.

.gitignore
bn222: We are sharing /tmp in the same way as we are sharing the cluster that is deployed. Only 1 test can run at 1 time because we are deploying and running on real hardware. This means that this kind of test already expects and manipulates external state.

`git clean -ndx` doesn't tear down a cluster.

thom311: ok.

PR #167, opened 2024-09-25
Fix vsp deploy

SalDaniele:
daemon: Crash if vsp init fails
    
    In the case that the vsp is not able to be started, the daemon should
    crash and retry.
    
    Currently there is a race condition when deploying the vsp from the
    operator where the vsp may fail to come up by the time "start" is
    called. This patch should address this race by retrying the daemon pod
    deployment.
    
    It makes sense to crash anyways, since the daemon on its own is useless
    without a running vsp, and we want to be able to easily see that an
    error has occured (i.e. daemon ends up in crashLoopBackoff if the vsp is
    never deployed).
    
    GrpcPlugin: Don't deploy vsp if image is not set
    
    Make sure we don't deploy vsp in the case WithVspImage is called with an
    empty vsp string.
    
    plugin: Use string constants for VSP image names

### DIFF
--- a/internal/daemon/plugin/vendorplugin.go
+++ b/internal/daemon/plugin/vendorplugin.go
@@ -21,9 +21,12 @@ import (
 //go:embed bindata/*
 var binData embed.FS
 
+const VspImageIntel string = "IntelVspImage"
+const VspImageMarvell string = "MarvellVspImage"
+
 var VspImages = []string{
-	"IntelVspImage",
-	"MarvellVspImage",
+	VspImageIntel,
+	VspImageMarvell,
 	// TODO: Add future supported vendor plugins here
 }
 
--- a/internal/daemon/plugin/vendorplugin.go
+++ b/internal/daemon/plugin/vendorplugin.go
@@ -106,9 +106,11 @@ func WithVspImage(template_vars map[string]string) func(*GrpcPlugin) {
 		vspImage := template_vars["VendorSpecificPluginImage"]
 		d.vspImage = vspImage
 		d.log.Info("VSP Image", "vspImage", d.vspImage)
-		err := render.ApplyAllFromBinData(d.log, "vsp-ds", template_vars, binData, d.k8sClient, nil, nil)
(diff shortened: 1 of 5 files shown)


thom311:
lgtm

thom311 (approved):

SalDaniele:
/override ci/prow/e2e-dpu-operator-tests

test passed in Jenkins (# 56)

internal/daemon/plugin/vendorplugin.go
```diff
@@ -106,9 +109,11 @@ func WithVspImage(template_vars map[string]string) func(*GrpcPlugin) {
 		vspImage := template_vars["VendorSpecificPluginImage"]
 		d.vspImage = vspImage
 		d.log.Info("VSP Image", "vspImage", d.vspImage)
-		err := render.ApplyAllFromBinData(d.log, "vsp-ds", template_vars, binData, d.k8sClient, nil, nil)
-		if err != nil {
-			d.log.Error(err, "Failed to start vendor plugin container", "vspImage", d.vspImage)
+		if vspImage != "" {
```
wizhaoredhat, line 112: After discussion with Sal.
This is quite confusing especially when the function name is called `WithVspImage`
We should call this code in a new function called `DeployVspImage` which is called when we create a new Grpc Plugin.

thom311: this `With*` function is a "Functional Options Pattern" in golang.

Me personally, I am dubious about whether this is really a great pattern, but it's at least a wide spread and common pattern.

In this case, the functions that return the callbacks are commonly called `WithSomething()`. `DeployVspImage` seems thus a bad name.

Feel free to open a patch to whatever you wish.

SalDaniele: PTAL at @168

SalDaniele: PTAL at @https://github.com/openshift/dpu-operator/pull/168

PR #168, opened 2024-09-27
Cleanup

SalDaniele:
plugin: Clean up vsp deployment code
    
    We should not deploy the vsp as a side effect of withVspImage. Rather we
    should explicitly call a function to handle deploy the Vsp.

    hack: add wa for IIC-364
    
    After IPU host reboot, the vsp init call fails to set an ip to
    communicate with the ACC. Restart the vsp/daemon to ensure this
    connection is up and ready to recieve cni requests.

### DIFF
--- a/hack/deploy_traffic_flow_tests.sh
+++ b/hack/deploy_traffic_flow_tests.sh
@@ -5,6 +5,13 @@ export KUBECONFIG=/root/kubeconfig.ocpcluster
 nodes=$(oc get nodes)
 export worker=$(echo "$nodes" | grep -oP '^worker-[^\s]*')
 
+
+# wa for https://issues.redhat.com/browse/IIC-364
+make undeploy
+make local-deploy
+oc create -f examples/host.yaml
+sleep 15 # Give times for Intel VSP to configure ip on <ipu-netdev>d3
+
 export KUBECONFIG=/root/kubeconfig.microshift
 nodes=$(oc get nodes)
 export acc=$(echo "$nodes" | grep -oP '^\d{3}-acc')
--- a/internal/daemon/plugin/vendorplugin.go
+++ b/internal/daemon/plugin/vendorplugin.go
@@ -64,19 +64,36 @@ type GrpcPlugin struct {
 	opiClient   opi.BridgePortServiceClient
 	nfclient    pb.NetworkFunctionServiceClient
 	dpuMode     bool
-	vspImage    string
+	vsp         VspTemplateVars
 	conn        *grpc.ClientConn
 	pathManager utils.PathManager
 }
 
-func CreateVspImageVars(vspImage string) map[string]string {
-	// All the CRs will be in the same namespace as the operator config
+func NewVspTemplateVars() VspTemplateVars {
+	return VspTemplateVars{
+		VendorSpecificPluginImage: "",
+		Namespace:                 "openshift-dpu-operator",
+		ImagePullPolicy:           "Always",
+		Command:                   "[ ]",
+		Args:                      "[ ]",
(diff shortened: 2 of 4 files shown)


internal/daemon/plugin/vendorplugin.go
```diff
@@ -131,8 +153,10 @@ func NewGrpcPlugin(dpuMode bool, client client.Client, opts ...func(*GrpcPlugin)
 		opt(gp)
 	}
 
-	if gp.vspImage == "" {
+	if gp.vsp.VendorSpecificPluginImage == "" {
 		gp.log.Info("WARNING: VSP Image not set, skipping vendor plugin container startup")
+	} else {
+		gp.deployVspDs()
```
thom311, line 159: I think the `if` check should move inside `deployVspDs()`.

thom311 (approved):
ltgm

internal/daemon/plugin/vendorplugin.go
SalDaniele: Yeah good point, updated

hack/deploy_traffic_flow_tests.sh
```diff
@@ -5,6 +5,14 @@ export KUBECONFIG=/root/kubeconfig.ocpcluster
 nodes=$(oc get nodes)
 export worker=$(echo "$nodes" | grep -oP '^worker-[^\s]*')
 
+
+# wa for https://issues.redhat.com/browse/IIC-364
+make undeploy
+podman push $(hostname):5000/intel_vsp:dev # Intel vsp is not multiarch, make sure intel_vsp:dev is x86_64
```
wizhaoredhat, line 11: How does this ensure that the image is x86_64 by pushing it again?

hack/deploy_traffic_flow_tests.sh
```diff
@@ -5,6 +5,14 @@ export KUBECONFIG=/root/kubeconfig.ocpcluster
...
+make local-deploy
+oc create -f examples/host.yaml
+sleep 15 # Give times for Intel VSP to configure ip on <ipu-netdev>d3
```
wizhaoredhat, line 14: How difficult would it be to check for an ip on <ipu-netdev>d3  instead of sleeping?

bn222: this is a wa, not worth it imo to remove the sleep

hack/deploy_traffic_flow_tests.sh
SalDaniele: We still can't cross compile Intel's VSP with podman, so we are building it on both sides when deploying the dpu operator. The images will have the same name currently, so this makes sure the x86 one is most recent.

Now that I am saying this, I should probably just push a patch to CDA to push these images as a buildah manifest so we don't need to worry about this

bn222: thanks

SalDaniele: Added in https://github.com/bn222/cluster-deployment-automation/pull/258, when this merges I will add an additional bump to CDA in this PR and remove this WA

SalDaniele: resolved

SalDaniele:
@bn222 PTAL, should be good to merge now

wizhaoredhat:
LGTM

SalDaniele:
passed on https://net-hw-ci-jenkins-csb-nst.apps.ocp-c1.prod.psi.redhat.com/view/IPU%20Testing/job/99_E2E_IPU_Deploy/118/parameters

PR #171, opened 2024-10-04
Operator sdk bump

bn222:


bn222:
@SalDaniele @wizhaoredhat ptal

wizhaoredhat:
@bn222 Is there a reason why we need to bump operator sdk?

bn222:
@wizhaoredhat bump includes controller runtime, which also needed to be bumped in prep of ghw bump

PR #172, opened 2024-10-07
Add sriov dp admission controller to dpu operator

vrindle:
SRIOV DP admission controller is needed in order to inject resource limits and requests for the dpu operator based on the specified network attachment definition in the pod yaml. This functionality exists in the SRIOV dp admission controller, so we need to enable it in the dpu operator (see repo here: https://github.com/openshift/sriov-dp-admission-controller). Link to jira for the scope of this: https://issues.redhat.com/browse/IIC-367


vrindle:
WIP (unable to run make-e2e-test locally). Fixing that before removing do-not-merge label.

internal/controller/bindata/webhook/auth.yaml
```diff
@@ -0,0 +1,163 @@
+# Copyright (c) 2019 Intel Corporation
```
bn222, line 1: ?

thom311:
what does `dp` in `sriov dp admission controller` stand for?

also, could you please describe the feature in the commit message and link to relevant sources (like the Jira issue that maybe contains additional information).

vrindle:
@thom311 ack. 

By the way... SRIOV dp admission controller is this repo here: https://github.com/openshift/sriov-dp-admission-controller

tools/tools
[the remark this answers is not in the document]
vrindle: ahh thanks for telling me that... the next commit I make I will make sure to delete :)

internal/controller/bindata/webhook/01_webhook.yaml
```diff
@@ -0,0 +1,28 @@
...
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
+webhooks:
+  - name: network-resources-injector-config.k8s.io
+    sideEffects: None
+    admissionReviewVersions: ["v1", "v1beta1"]
+    clientConfig:
+      service:
+        name: network-resources-injector-service
+        namespace: "openshift-dpu-operator"
```
wizhaoredhat, line 16: namespace: {{.Namespace}}

internal/controller/bindata/webhook/01_webhook.yaml
```diff
@@ -0,0 +1,28 @@
...
+        path: "/mutate"
+    namespaceSelector:
+      matchExpressions:
+        - key: "kubernetes.io/metadata.name"
+          operator: "NotIn"
+          values:
+            - "openshift-dpu-operator"
```
wizhaoredhat, line 23: - {{.Namespace}}

internal/controller/bindata/webhook/14_server.yaml
```diff
@@ -0,0 +1,63 @@
...
+      app: network-resources-injector
+  template:
+    metadata:
+      labels:
+        app: network-resources-injector
+    spec:
+      serviceAccount: network-resources-injector-sa
+      containers:
+      - name: webhook-server
+        image: quay.io/vpunj/sriov-dp-admission-controller:latest
```
wizhaoredhat, line 20: Please change to quay.io/openshift/sriov-dp-admission-controller:latest

wizhaoredhat:
Tests are needed.

vrindle:
@wizhaoredhat I talked with @bn222 last week about the tests... we discussed that removing the workaround from the sriov dp admission controller would be the way to test this PR... that PR is in review.

internal/controller/bindata/webhook/14_server.yaml
bn222: can you add `origin-` to that url? quay.io/openshift/*origin-*sriov-dp-admission-controller

SalDaniele: Should this be dynamically set? Similar to how we set the daemon / vsp images, so we can [point to different images for development](https://github.com/openshift/dpu-operator/blob/main/config/dev/local-images-template.yaml), and have all production images specified [in a single place](https://github.com/openshift/dpu-operator/blob/main/config/manager/manager.yaml)?

vrindle:
/hold
As per @SalDaniele message CI failure is due to https://github.com/openshift/dpu-operator/pull/302 not being in

SalDaniele:
/test make-e2e-test 

Failed due to registry.ci images being down

SalDaniele:
@thom311 PTAL when you have a chance, would like to get your input before merging

cmd/main.go
```diff
@@ -94,7 +94,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	b := controller.NewDpuOperatorConfigReconciler(mgr.GetClient(), mgr.GetScheme(), dpuDaemonImage, vspImages)
+	webhookImage := os.Getenv("WEBHOOK_IMAGE")
+	if webhookImage == "" {
```
SalDaniele, line 98: So if I set "ENABLE_WEBHOOKS to false, I will still get this particular webhook: https://github.com/openshift/dpu-operator/blob/main/cmd/main.go#L114

Should we converge the logic to deploy these?

Maybe we need a more descriptive name in general to differentiate these webhooks

SalDaniele (commented):
Overall LGTM.

Can we add a test to validate the webhook is functioning properly?

vrindle:
@SalDaniele FYI: the test is added if pod2pod is passing without the workaround, the webhook is passing as expected...

cmd/main.go
```diff
@@ -94,7 +94,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	b := controller.NewDpuOperatorConfigReconciler(mgr.GetClient(), mgr.GetScheme(), dpuDaemonImage, vspImages)
+	webhookImage := os.Getenv("WEBHOOK_IMAGE")
```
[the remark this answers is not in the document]
vrindle: This is the way that the sriov-network-operator deploys it (by scaling up a pod to handle it). Both are valid ways of deploying the webhook.

cmd/main.go
vrindle: You will... but this webhook is specified as a pod called the network-resources-injctor when deployed.

SalDaniele: I mean from the perspective of someone reading the code

SalDaniele:
Discussed offline with Vrinda, makes sense to me to remove the workaround. Adding a test where we validate the resource is being injected still seems like a reasonable thing to do, but if make-e2e is passing, the code is working as expected.

hack/traffic_flow_tests.sh
```diff
@@ -27,4 +27,5 @@ temp_file=$(mktemp)
 
 envsubst < ../hack/cluster-configs/ocp-tft-config.yaml > $temp_file
 
+TFT_MANIFESTS_OVERRIDE=/root/dpu-operator/hack/cluster-configs/ocp_traffic_flow_manifests
```
thom311, line 30: I think this should be a relative path, so that it works with any checkout. Also note how the Jenkins systems tend to checkout `/home/jenkins/workspace/$PROJECT/dpu-operator/`.

Also, I think this needs an `export TFT_MANIFESTS_OVERRIDE=...` to take effect.

Also, the variables is called `TFT_MANIFESTS_OVERRIDES`. The README is wrong, there needs to be a trailing "S" (I will fix the k8s-tft README).

In the logfile of tft, you will see which path is used. Would be good to confirm that it works.

hack/cluster-configs/ocp_traffic_flow_manifests/sriov-pod.yaml.j2
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
thom311, line 1: the project was renamed from `ocp-traffic-flow-tests` to `kubernetes-traffic-flow-tests`. It seems better to not stick to the old name.

I realize, that there is already a pre-existing `ocp-tft-config.yaml`. Let's ignore the inconsistency of that name for now.

But the new directory, I think it should just be called `traffic_flow_tests_manifests/`, or `tft_manifests/` or `traffic_flow_tests/manifests/` or `tft/manifests/`. I like `tft/manifests/` the most.

Note that we also have `hack/traffic_flow_tests.sh`, so it's already the case that not all our files related to k8s-tft are named with something "ocp".

thom311 (commented):
This looks reasonable to me, as we follow sriov-network-operator's webhook.

Only minor comments about tft script.

PR #175, opened 2024-10-10
fix deprecated field and rename and move e2e files

bn222:


hack/deploy_traffic_flow_tests.sh
```diff
@@ -9,7 +9,7 @@ export KUBECONFIG=/root/kubeconfig.microshift
 nodes=$(oc get nodes)
 export acc=$(echo "$nodes" | grep -oP '^\d{3}-acc')
 
-envsubst < ../cluster_configs/config.yaml > tft_config.yaml
+envsubst < cluster-configs/config.yaml > tft_config.yaml
```
thom311, line 12: as you already rename things, could you rename this file?

The other `config*.yaml` files in this directory are configs for CDA deploy.
This file is the configuration file for ocp-tft, it's a different format and a different purpose.

I would choose `ocp-tft-config.yaml`.

PR #182, opened 2024-10-18
Enable traffic flow tests

bn222:


bn222:
/override e2e-dpu-operator-tests 
Known issue with tft

PR #186, opened 2024-10-18
Add make_fast_e2e_test target to dpu operator

vrindle:


vrindle:
@bn222

hack/deploy_fast.sh
```diff
@@ -0,0 +1,41 @@
...
+    local KUBECONFIG_PATH=$1
+
+    export KUBECONFIG=$KUBECONFIG_PATH
+    
+    if oc get nodes; then
+        return 0
+    else
+        return 1
+    fi
+}
```
bn222, line 13: We don't know if this is the right cluster. This is ok for now, but it's not really fool-proof.

hack/deploy_fast.sh
```diff
@@ -0,0 +1,41 @@
...
+check_ocp "/root/kubeconfig.ocpcluster"
+result_ocp=$?
+
+check_ocp "/root/kubeconfig.microshift"
+result_microshift=$?
+
+# Check if both OCP and MicroShift checks succeeded
+if [ $result_ocp -eq 0 ] && [ $result_microshift -eq 0 ]; then
+    echo "Both OCP and MicroShift cluster checks succeeded!"
+    deploy_local_dpu_operator
```
bn222, line 36: where is the undeploy part?

vrindle: We do not need it... here we are deploying dpu operator using 'post' in CDA. When deploying the dpu operator, CDA already takes care of the idempotent issue by doing a `make-undeploy` prior to deploying. See [here](https://github.com/bn222/cluster-deployment-automation/blob/65be967c565f97cf87322f6858b5f18f3b3efb16/extraConfigDpu.py#L127). By using the post config we can take advantage of this feature without specifying it in this script.

bn222: ok sgtm

hack/deploy_fast.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+check_ocp() {
+    local KUBECONFIG_PATH=$1
+
+    export KUBECONFIG=$KUBECONFIG_PATH
+    
+    if oc get nodes; then
```
SalDaniele, line 8: Fine with merging this initially, but there is a (pretty good) chance we try to deploy this test on a broken cluster, which will quickly add up with false fast-test failures.

Maybe a good strategy is:
- Add improved assessments of cluster health (i.e. nodes are ready, all cluster operators are ready and working)
- If I try to run a fast-test on a broken cluster, it should run as a slow test (redeploy the whole thing).

False failures might increase the burden on the contributor to track down whether their code broken something, or whether the CI is broken.

vrindle: totally agree.... I have to figure out ways I can add cluster-health-checks to this.

SalDaniele: Or we could dedicate a server to fast-tests which will:
- redeploy lastest OCP nightly on a schedule
- if failing cluster state is detected, re-install OCP

vrindle: @SalDaniele I discussed the idea of having a server dedicated to fast tests with @bn222. The issue with that is that we want our CI running (having 1 server can create a potential issue where fast tests are done, but slow tests are in queue - we want to avoid that scenario by pooling them together)

hack/deploy_fast.sh
SalDaniele: In the case of the IPU there are still artifacts that are not cleaned up by "make-undeploy", specifically the p4 container. We need to make this idempotent in CDA:

```
[root@238-acc ~]# podman ps
CONTAINER ID  IMAGE                                                                         COMMAND     CREATED            STATUS            PORTS       NAMES
7e2630fc7e0b  wsfd-advnetlab239.anl.eng.bos2.dc.redhat.com:5000/intel-ipu-p4-sdk:10-9-2024              About an hour ago  Up About an hour              suspicious_ishizaka
7f2e5b924393  wsfd-advnetlab239.anl.eng.bos2.dc.redhat.com:5000/intel-ipu-p4-sdk:10-9-2024              43 minutes ago     Up 43 minutes                 reverent_kapitsa
a83bc282138a  wsfd-advnetlab239.anl.eng.bos2.dc.redhat.com:5000/intel-ipu-p4-sdk:10-9-2024              25 minutes ago     Up 25 minutes                 jovial_hodgkin
392b5faefce5  wsfd-advnetlab239.anl.eng.bos2.dc.redhat.com:5000/intel-ipu-p4-sdk:10-9-2024              8 minutes ago      Up 8 minutes                  nice_banzai
```

PR #190, opened 2024-10-21
Use mktemp when running vendor check

bn222:


bn222:
/override ci/prow/make-fast-e2e-test

This touches only the Makefile that won't be used by e2e test

PR #192, opened 2024-10-21
Add Host Network Attachment Definition

SamD2021:
# TL;DR
This PR makes some changes to how we handle Network Attachment Definitions (NAD) in the DPU Operator. The goal is to centralize the NAD setup, and automate the deployment of the NAD for the Host. I'm also hoping to discuss if these changes should be moved to the daemon so we can support auto mode as well.

### Host NAD YAML:

Added a manifest for Host NAD to centralize its configuration in the operator itself.
The namespace follows what we use in Cluster Deployment Automation (CDA), so the default resources get nad properly.

### NAD Reconciliation:

Changed the check for NAD to ensure both Host and DPU mode have deployed its NADs while skipping it for Auto mode.

### Auto Mode Handling:

If the configuration is set to auto, the operator will skip NAD deployment since the controller can’t detect the hardware in this mode. This keeps things cleaner and avoids unnecessary errors.

## Why These Changes Matter
By centralizing the Host NAD setup and improving how we handle reconciliation, this makes the operator more reliable and lessens the burden of configuration from the user. I also hope to have a discussion about whether NAD management should be the job of the controller or the daemon, since the latter is the one that can auto detect the hardware.

### DIFF
--- a/internal/controller/bindata/networkfn-nad-host/00.networkfun-nad.yaml
+++ b/internal/controller/bindata/networkfn-nad-host/00.networkfun-nad.yaml
@@ -1,0 +1,17 @@
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: sriov-net1
+  namespace: default
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: {{.ResourceName}}
+spec:
+  config: '{
+    "type": "dpu-cni",
+    "cniVersion": "0.4.0",
+    "name": "dpu-cni",
+    "ipam": {
+      "type": "host-local",
+      "subnet": "10.56.217.0/24"
+    }
+    }'
--- a/internal/controller/dpuoperatorconfig_controller.go
+++ b/internal/controller/dpuoperatorconfig_controller.go
@@ -141,7 +141,16 @@ func (r *DpuOperatorConfigReconciler) ensureDpuDeamonSet(ctx context.Context, cf
 func (r *DpuOperatorConfigReconciler) ensureNetworkFunctioNAD(ctx context.Context, cfg *configv1.DpuOperatorConfig) error {
 	logger := log.FromContext(ctx)
 	logger.Info("Create the Network Function NAD")
-	return r.createAndApplyAllFromBinData(logger, "networkfn-nad", cfg)
+	nadFile := ""
+	switch cfg.Spec.Mode {
+	case "dpu":
+		nadFile = "networkfn-nad"
+	case "host":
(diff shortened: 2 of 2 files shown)


internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -99,14 +99,13 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 
-	if dpuOperatorConfig.Spec.Mode == "dpu" {
+	if dpuOperatorConfig.Spec.Mode != "auto" {
```
SalDaniele, line 102: @bn222 @wizhaoredhat discussed this with Sam. Should we move creating the net-attach-def for the host to the daemon as well?

The issue being that the top level operator does not have access to the platform info to detect whether it should create a NAD for the host-side or dpu-side, other than the dpu operator config spec.

However I don't think we can rely on this spec in the case we auto detect mode.

SalDaniele:
As part of this PR I think we should remove creating the nad in CDA, and add a bump to CDA as a commit to make sure e2e tests are still working.

internal/controller/bindata/networkfn-nad-host/00.networkfun-nad.yaml
```diff
@@ -0,0 +1,17 @@
...
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: {{.ResourceName}}
+spec:
+  config: '{
+    "type": "dpu-cni",
+    "cniVersion": "0.4.0",
+    "name": "dpu-cni",
+    "ipam": {
+      "type": "host-local",
+      "subnet": "10.56.217.0/24"
```
wizhaoredhat, line 15: I don't know how useful this change is for the host is.

1. As you noticed "ipam" can be configured by the nad. The IP Address "10.56.217.0/24"" is not relevant for most if not all customer setups. Yes it is relevant for CDA, but we aren't building this for CDA.
2. Also  there are different types of IPAMs 
3. dpu-cni could configure parameters like MTU on the host.
4. In short, the reason why I only did the NAD for the DPU is that it is assumed the DPU's Network Functions will be created by the operator only. But the host NAD would be consumed by workload pods that are outside our control.

bn222: host NAD will be created by the operator as a consequence of creating a DpuNetwork CR

wizhaoredhat: This means that the DpuNetwork CR must contain IPAM settings for the NAD in the future.

PR #194, opened 2024-10-25
[th/hack-scripts-cleanup] cleanup hack scripts

thom311:
- hack: make scripts executable and add shebang
- hack: use "set -e" to fail scripts on error
- Makefile: directly call scripts instead of using variables

### DIFF
--- a/hack/deploy_fast.sh
+++ b/hack/deploy_fast.sh
@@ -1,10 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 check_ocp() {
     local KUBECONFIG_PATH=$1
 
     export KUBECONFIG=$KUBECONFIG_PATH
-    
+
     if oc get nodes; then
         return 0
     else
@@ -38,4 +38,3 @@ else
     echo "Cluster config is not proper... deploying cluster and dpu operator"
     deploy_cluster_and_dpu_operator
 fi
-
--- a/hack/deploy_fast.sh
+++ b/hack/deploy_fast.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 check_ocp() {
     local KUBECONFIG_PATH=$1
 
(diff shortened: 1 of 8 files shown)


hack/prepare.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 git submodule init
 git submodule update
```
bn222, line 6: Can you add `|| true` for the rest of the script after the git submodule init? Otherwise `set -e` will not work for my workflow anymore.

thom311: To all of them? That seems not right, at that point, I would `set +e`.

Which commands are known/expected to fail?

bn222: I need git submodule init and update separately. Can we refactor instead? prepare-repo and have a separate target for that? Then, we can keep `set -e`

thom311: This patch is about using `set -e` and not silently ignoring errors, it does not question which commands to call.

Usually, `git submodule init` and `git submodule update` commands do not fail, if the directory is already initialized/updated. Could you explain the case where they are known to fail?

thom311: From your comments I don't understand what the problem is (or what you suggest as solution).

This patch is about scripts not ignoring errors. Nothing more.

If you want to move code around or remove it entirely, then please either send a patch or elaborate on the problem.

bn222: Can you rebase on top of https://github.com/openshift/dpu-operator/pull/198 ?

thom311: I will do (once #198 is merged). Setting PR back to Draft for now (due to that).

hack/deploy_fast.sh
```diff
@@ -1,10 +1,12 @@
...
+
+set -e
 
 check_ocp() {
     local KUBECONFIG_PATH=$1
 
     export KUBECONFIG=$KUBECONFIG_PATH
-    
+
     if oc get nodes; then
```
SalDaniele, line 10: I believe this patch is failing the fast test because we set -e on this script: https://net-hw-ci-jenkins-csb-nst.apps.ocp-c1.prod.psi.redhat.com/view/IPU%20Testing/job/99_FAST_E2E_IPU_Deploy/75/console

It is possible check_ocp() runs oc get nodes which will fail, we dont' want to bail out in this case, but rather redeploy the cluster.

thom311: right. That is now fixed.

PR #198, opened 2024-10-28
separate prepare script

bn222:


hack/prepare-venv.sh
```diff
@@ -1,8 +1,5 @@
 #!/usr/bin/env bash
 
-git submodule init
-git submodule update
-
 cd cluster-deployment-automation
 rm -rf /tmp/cda-venv
 python3.11 -m venv /tmp/cda-venv
```
thom311, line 5: The purpose of this script is not obvious. It may be fine, but neither the commit message nor the PR description describes why this is done. Yes, it splits some step of the script to another script (an obvious fact pointed out by the commit message). But why? And why should this not be reverted next month? The reason why this is done should be recorded in the git commit message.

bn222: This script's execution depends on the env it runs in. By allowing it to run separately, it can be run in more envs even without git.

PR #200, opened 2024-10-28
Proper device handler

SalDaniele:
sriovDevHandler: Call out to vsp for dev handling
    
    Previously, the Sriov Device Handler would manually set num vfs /
    retrieve the pci device information since this api was not provided by
    the vsp.
    
    This has since been added [1]. This function should be offloaded to the
    vsp.
    
    Follow up to [2].
    
    [1] https://github.com/intel/ipu-opi-plugins/pull/223/files#diff-8c7117bca54284b14fe9e0bc34ae12dc71015bf5f97bad23b66ff8a5b0dac50a
    [2] https://github.com/openshift/dpu-operator/pull/124/commits/82693320263591feda2c535c085fb615edad4732#diff-0fbf78ae52181d6f4802d984f45af934e1dcc5f8f2c9c044b720c83493dcb4e6


SalDaniele:
@bn222 PTAL this is needed to complete the tasks for M1

PR #202, opened 2024-10-31
Deploy in parallel

bn222:


hack/deploy_fast.sh
```diff
@@ -15,15 +15,13 @@ check_ocp() {
...
-   bash hack/deploy_traffic_flow_tests.sh
+    bash hack/both.sh
+    bash hack/deploy_traffic_flow_tests.sh
 }
 
 deploy_local_dpu_operator() {
-   bash hack/ipu_host_deploy_post.sh
-   bash hack/ipu_deploy_post.sh
-   bash hack/deploy_traffic_flow_tests.sh
+    bash hack/both.sh
```
SalDaniele, line 23: shouldnt this only run the post steps?

hack/deploy_fast.sh
```diff
@@ -40,3 +37,5 @@ else
     echo "Cluster config is not proper... deploying cluster and dpu operator"
     deploy_cluster_and_dpu_operator
 fi
+
+bash hack/deploy_traffic_flow_tests.sh
```
thom311, line 41: the script should be made executable and have a shebang. Thereby it chooses its own interpreter, and you don't need to choose `bash` for it.

hack/both.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
```
thom311, line 2: as a matter of principle, scripts should run with `set -e` and errors should be handled (or left unhandled an be fatal).

well, if you insist on handling errors below, then indeed a `set -e` may make it a bit more cumbersome. But such error handling `if [ $rc -eq 0 ]` seems unnecessary anyway. If you want to make the script easier to debug, maybe use `set -ex`.

Makefile
```diff
@@ -85,13 +85,9 @@ ifeq ($(SUBMODULES), true)
...
-.PHONY: ipu_host
-ipu_host: prepare-e2e-test
-	./hack/ipu_host_deploy.sh
-
-.PHONY: ipu_deploy
-ipu_deploy: prepare-e2e-test
-	./hack/ipu_deploy.sh
+.PHONY: deploy_clusters
+deploy_clusters: prepare-e2e-test
+	bash hack/both.sh
```
thom311, line 90: why adding all those make file targets? The script can be called directly. If it has a good name it's also discoverable.

PR #204, opened 2024-11-05
[th/make-generate-tidy] makefile: run `go mod tidy` during `make generate`

thom311:
We have multiple go.mod files, this does not seem to agree with operator-sdk. operator-sdk will generate files and call `make generate`. However, that will then fail because we didn't call `go mod tidy` for api/ directory.

Note that compared to sriov-network-operator, in dpu-operator we have several go.mod files. That does not seem to agree with operator-sdk, or at least, something is not hooked up right.

Hack `make generate` to call `go mod tidy` first.

Otherwise:
```
  $ git checkout -B C origin/main && \
    git reset --hard HEAD && \
    git clean -fdx && \
    make operator-sdk && \
    ./bin/operator-sdk create webhook --group config --version v1 --kind DpuOperatorConfig --defaulting --programmatic-validation
  branch 'C' set up to track 'origin/main'.
  Reset branch 'C'
  Your branch is up to date with 'origin/main'.
  HEAD is now at e58a9c9c830a Merge pull request #194 from thom311/th/hack-scripts-cleanup
  INFO[0000] Writing kustomize manifests for you to edit...
  ERRO[0000] Unable to find the target #- path: manager_webhook_patch.yaml to uncomment in the file config/default/kustomization.yaml.
  INFO[0000] Writing scaffold for you to edit...
  INFO[0000] api/v1/dpuoperatorconfig_webhook.go
  INFO[0000] api/v1/dpuoperatorconfig_webhook_test.go
  INFO[0000] api/v1/webhook_suite_test.go
  INFO[0000] Update dependencies:
  $ go mod tidy
  INFO[0000] Running make:
  $ make generate
  test -s /data/src/dpu-operator/bin/controller-gen && /data/src/dpu-operator/bin/controller-gen --version | grep -q v0.15.0 || \
  GOBIN=/data/src/dpu-operator/bin GOFLAGS='' go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.15.0
  GOFLAGS='' /data/src/dpu-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
  Error: load packages in root "/data/src/dpu-operator/api": err: exit status 1: stderr: go: updates to go.mod needed; to update it:
          go mod tidy
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,15 @@ prow-ci-manifests-check: manifests
 		exit 1; \
 	fi
 
+.PHONY: vendor
+vendor:
+	for d in . dpu-api api tools ; do \
+		if [ "$$d" = . ] ; then \
+			(cd $$d && go mod vendor) || exit $$? ; \
+		fi ; \
+		(cd $$d && go mod tidy) || exit $$? ; \
+	done
+
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	GOFLAGS='' $(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
--- a/Makefile
+++ b/Makefile
@@ -152,24 +152,13 @@ vendor:
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	GOFLAGS='' $(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
-TMP_DIR := $(shell mktemp -d)
 .PHONY: generate-check
 generate-check: controller-gen
-	rm -rf $(TMP_DIR)
-	mkdir -p $(TMP_DIR)
-	cp -r . $(TMP_DIR)
-	cd $(TMP_DIR) && make generate
-	diff -r . $(TMP_DIR) || exit 1
-	rm -rf $(TMP_DIR)
+	./scripts/check-gittree-for-diff.sh make generate
 
 .PHONY: vendor-check
 vendor-check:
-	rm -rf $(TMP_DIR)
-	mkdir -p $(TMP_DIR)
-	cp -r . $(TMP_DIR)
-	cd $(TMP_DIR) && go mod vendor && go mod tidy
-	diff -r . $(TMP_DIR) || exit 1
-	rm -rf $(TMP_DIR)
+	./scripts/check-gittree-for-diff.sh make vendor
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
--- a/scripts/check-gittree-for-diff.sh
+++ b/scripts/check-gittree-for-diff.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/bash
+
+die() {
+    printf "%s\n" "$*"
+    exit 1
+}
+
+set -e
+
+test $# -gt 0 || die "Usage: $0 CMD..."
+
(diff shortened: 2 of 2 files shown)


thom311:
> Doesn't this behave the same way if we didn't have a go.mod?

This patch has only an effect, when `make generate` is called.

I don't understand what k8s' `make generate` is supposed to do. So the answer depends on that.

bn222:
I don't think we want to mix make generate with go mod tidy. I think go mod tidy should be run separately.

thom311:
> I don't think we want to mix make generate with go mod tidy. I think go mod tidy should be run separately.

There is a reproducer in the commit message that you can just run.

How do you suggest to solve this differently? Especially seeing that `make generate` is not run by me directly, but by `operator-sdk` indirectly?

bn222:
It think the right place is `make vendor` , and we also need to extend `make vendor-check`

Makefile
```diff
@@ -151,29 +151,24 @@ tidy:
...
 
 .PHONY: vendor-check
 vendor-check:
-	rm -rf $(TMP_DIR)
-	mkdir -p $(TMP_DIR)
-	cp -r . $(TMP_DIR)
-	cd $(TMP_DIR) && go mod vendor && go mod tidy
-	diff -r . $(TMP_DIR) || exit 1
-	rm -rf $(TMP_DIR)
+	./scripts/check-gittree-for-diff.sh make vendor
```
bn222, line 171: this doesn't check go mod tidy anymore?

bn222: We need vendor-check to make sure that we properly vendor and run go mod tidy in the right directories.

thom311: `make vendor` includes `go mod tidy`.

However, there is a problem, namely that calling `make vendor` generates files files `{api,dpu-api}/vendor/`.

Pushed a fix.

bn222: This is a bit confusing, I expect we do vendor in `.` and tidy in all directories where needed.

Makefile
```diff
@@ -143,7 +143,16 @@ prow-ci-manifests-check: manifests
...
+
+.PHONY: tidy
+tidy:
+	for d in $(GO_MOD_DIRS) ; do \
+		(cd $$d && go mod tidy) || exit $$? ; \
+	done
+
 .PHONY: generate
+generate: tidy
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
```
bn222, line 165: can we have these on the same line? Also, can we make it depend on vendor and remove tidy target?

thom311: ok...

there is a minor problem that running `make vendor` before the `make generate` step, is that that can result in something, which still gets changed by another `make vendor`. But that is only in the case where `make generate` even does anything. So, in some cases, you may still want to run `make generate && make vendor`, although that really runs `make vendor && make generate && make vendor`. But I think that's OK (and not the common case).

thom311:
Note that the `build` target also includes the `generate` target (`build` is also the `default` target). That means, a plain `make` or `make build` now also runs `go mod tidy`. That seems wrong that a build tries to generate/tidy the source files.

But I think the problem of that is that it's wrong that `build` does a `generate`. The problem is not that `generate` should not do a `go mod tidy`.

thom311:
See similar comment from @bn222 [here](https://github.com/openshift/dpu-operator/pull/213#discussion_r1847798566):

> This prevents separate commits with vendor

That part is now dropped. `make generate` will no longer also depend on `vendor` target. Which defeats most of the purpose of why this commit was done. However, that part will temporarily come back in #213.

PR #206, opened 2024-11-08
[th/revert-device-ids] Revert "deviceHandler: Handle both net dev and pci IDs"

thom311:
This breaks the unit test

    [FAIL] DPU side maanger Device Plugin [It] Should allocate devices
    /tmp/do/internal/daemon/dpusidemanager_test.go:44

with

    ERROR   DevicePlugin    Failed to get Devices   {"error": "Failed to normalize device ens5f3 from GetDevice request: failed to get PCI address for netdev ens5f3: failed to find PCI address for net device ens5f3: lstat /sys/class/net/ens5f3: no such file or directory"}

The Mock plugin just returns some interface names that don't actually exist and don't have a PCI address. Trying to resolve them as a PCI address does not work. Especially it does not work if that is done by dpuDeviceHandler.GetDevices() as this method is not mocked. To make unit tests work in that case, we would have to mock
dpuDeviceHandler.GetDevices(). But that seems wrong. Instead, this function should not require mocking and just treat the device ID as an opaque identifier.

 > As a result we need to handle the case netdevs are provided. Internally
 > the device plugin expects all devies in PCI address format

Then the device plugin needs to handle that. The places that need to know something about the device (like the PCI address) need to resolve the opaque ID to whatever they need. Possibly by asking the VSP to resolve the ID. That part maybe also needs to be mocked for unit tests.

Also, the patch gives meaning to "pluginapi.Device{ID: devPciId}" without renaming the attribute "ID". That is confusing.

Most importantly. Whatever the solution, revert for now, because this breaks unit tests.

This reverts commit f85129fdac1d18c4d89cccf443e4104b31e1fdea.

### DIFF
--- a/dpu-cni/pkgs/sriovutils/sriovutils.go
+++ b/dpu-cni/pkgs/sriovutils/sriovutils.go
@@ -49,17 +49,6 @@ func EnableArpAndNdiscNotify(ifName string) error {
 	return nil
 }
 
-// GetPciFromNetDev takes in a network device name and returns its PCI address
-func GetPciFromNetDev(ifName string) (string, error) {
-	netDevPath := filepath.Join(NetDirectory, ifName, "device")
-	pciAddr, err := filepath.EvalSymlinks(netDevPath)
-	if err != nil {
-		return "", fmt.Errorf("failed to find PCI address for net device %s: %v", ifName, err)
-	}
-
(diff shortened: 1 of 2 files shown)


bn222:
/hold

This functionality is needed and we need to properly fix the issue instead of reverting it.

PR #209, opened 2024-11-15
Kubecon

SalDaniele:


internal/daemon/device-handler/dpu-device-handler/dpudevicehandler.go
```diff
@@ -57,7 +57,13 @@ func (d *dpuDeviceHandler) GetDevices() (*dp.DeviceList, error) {
 
 	devices := make(dp.DeviceList)
 
+	// TODO: We need to properly enforce API boundaries at the VSP level. The host side requires pci-addresses when handling devices, however the dpu side requires a higher level of abstraction. For now, just enforce PCI addresses for device ID on the host only.
 	for _, device := range Devices.Devices {
+		if d.dpuMode {
```
wizhaoredhat, line 62: Shouldn't this compare with the Host mode?

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
```diff
@@ -25,19 +25,28 @@ spec:
         command: {{.Command}}
         args: {{.Args}}
         volumeMounts:
-        - name: vendor-plugin-sock
-          mountPath: /var/run/dpu-daemon/
-        - name: host-opt
-          mountPath: /opt/p4
+        - mountPath: /var/run/
+          name: vendor-plugin-sock
+        - mountPath: /opt/p4/p4-cp-nws/var
```
wizhaoredhat, line 30: Shouldn't /opt/p4 be good?

SalDaniele: Yeah I think so,  just opted for the more specific path, I could update this to /opt/p4

internal/daemon/device-handler/dpu-device-handler/dpudevicehandler.go
```diff
@@ -57,7 +57,13 @@ func (d *dpuDeviceHandler) GetDevices() (*dp.DeviceList, error) {
...
+			devices[device.ID] = pluginapi.Device{ID: device.ID, Health: pluginapi.Healthy}
+			continue
+		}
+
 		devPciId, err := normalizeDeviceToPci(device.ID)
```
thom311, line 67: does this fix the unit tests? That is, do the existing unit tests all run in DPU mode?

If not, either the test needs to be disabled/deleted or adjusted/fixed.

SalDaniele: This fixes the unit test incidentally, the intent here was just to fix a bug with the Intel IPU vendor plugin (IPU vsp provides devices by netdev, they all share the same PCI address so this would break).

We only use the mockVsp in dpu mode currently, so the test should pass on all machines now. If we were to implement the mockVsp in host mode, we would need to provide pci addresses, or else mock the code elsewhere to avoid failing the test if the netdev does not exist on the host it is running on.

I believe Balazs is working on a more systematic change to refactor the code's dependency on the hardware its running on, so perhaps this would address fixing this test case if we want to mock the vsp in host mode in the future.

As for the actual behavior of the code in a real deployment, this is not a good fix. normalizeDeviceToPci() should really be enforcePciAddress() or something, and the change needs to be made in the vendors code to use netdevs on the host side (since we determined this is what we want this API to be moving forward). But I do not want to have the code-base be broken until Intel fixes this, which is why I added this hack for now.

internal/daemon/device-handler/dpu-device-handler/dpudevicehandler.go
wizhaoredhat: I see, got it now.

PR #213, opened 2024-11-18
[th/validating-webhook] add validating admission webhook to check DpuOperatorConfig

thom311:
- use `operator-sdk create webhook` to add the scaffolding for a validating admission webhook
- implement checks for `DpuOperatorConfig`
  - enforce name to be `dpu-operator-config`
  - check mode for valid values

https://issues.redhat.com/browse/IIC-317

Missing: run the webhook in the kind cluster. As such, the tests don't cover this (though, there is a unit test that checks the validation code a bit). Maybe `make test` also need to run the admission webhook (if possible).


Makefile
```diff
@@ -143,8 +143,17 @@ prow-ci-manifests-check: manifests
...
+	for d in . dpu-api api tools ; do \
+		if [ "$$d" = . ] ; then \
+			(cd $$d && go mod vendor) || exit $$? ; \
+		fi ; \
+		(cd $$d && go mod tidy) || exit $$? ; \
+	done
+
 .PHONY: generate
-generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+generate: vendor controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
```
bn222, line 156: This prevents separate commits with vendor

thom311: how do you suggest to solve this?

The commit message explains the problem, and also has a reproducer that you can just run.

also, this commit is part of https://github.com/openshift/dpu-operator/pull/204 , which is why this PR is still a Draft.

wizhaoredhat:
This looks good! Very clean for implementing the webhook. The issues you had before, how did you solve them? What were the issues such that we know what can go wrong.

thom311:
> This looks good! Very clean for implementing the webhook. The issues you had before, how did you solve them? What were the issues such that we know what can go wrong.

The main problem is that I didn't understand how to correctly make TLS certificates for the webhook work. 

One approach is to uncomment the `[CERTMANAGER]` parts. That took me a while to understand how to get that to work. In particular, we must install the `cert-manager` operator, which is available in Openshift but not enabled by default. That is a problem for using it in dpu-operator.

Compared to sriov-network-operator, that seems implemented mostly by hand (not using operator-sdk?). So what is done there does not directly translate and spread out over many commits. It was not easy to understand or follow. It supports both cert-manager and service.beta.openshift.io.

For this branch, only service.beta.openshift.io is implemented. That seems sufficient to me, it also seems to work on Microshift.

Makefile
thom311: discussion moved to https://github.com/openshift/dpu-operator/pull/204#issuecomment-2491456950

PROJECT
```diff
@@ -21,7 +21,6 @@ resources:
   path: github.com/openshift/dpu-operator/api/v1
```
bn222, line 21: can you fix having to manually do `go get github.com/onsi/ginkgo/v2@v2.20.2`

thom311: the upgrade is done to avoid some failure in `make test`. Is there a problem, or what's the use of staying with an old version (and a different version from the toplevel `go.mod`)?

thom311: without `(cd api/ && go get github.com/onsi/ginkgo/v2@v2.20.2)`, a subsequent `make test` gives a warning:

```
$ make test
GOFLAGS='' /data/src/dpu-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
for d in . dpu-api api tools ; do \
        if [ "$d" = . ] ; then \
                (cd $d && go mod vendor) || exit $? ; \
        fi ; \
        (cd $d && go mod tidy) || exit $? ; \
done
GOFLAGS='' /data/src/dpu-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
test -s /data/src/dpu-operator/bin/setup-envtest || GOBIN=/data/src/dpu-operator/bin GOFLAGS='' go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.16
test -s /data/src/dpu-operator/bin/ginkgo && /data/src/dpu-operator/bin/ginkgo version | grep -q v2.20.2 || \
GOBIN=/data/src/dpu-operator/bin GOFLAGS='' go install github.com/onsi/ginkgo/v2/ginkgo@v2.20.2
FAST_TEST=false KUBEBUILDER_ASSETS="/data/src/dpu-operator/bin/k8s/1.27.1-linux-amd64" /data/src/dpu-operator/bin/ginkgo --repeat 4  ./... -coverprofile cover.out
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.20.2
  Mismatched package versions found:
    2.14.0 used by v1

  Ginkgo will continue to attempt to run but you may see errors (including flag
  parsing errors) and should either update your go.mod or your version of the
  Ginkgo CLI to match.

  To install the matching version of the CLI run
    go install github.com/onsi/ginkgo/v2/ginkgo
  from a path that contains a go.mod file.  Alternatively you can use
    go run github.com/onsi/ginkgo/v2/ginkgo
  from a path that contains a go.mod file to invoke the matching version of the
  Ginkgo CLI.

  If you are attempting to test multiple packages that each have a different
  version of the Ginkgo library with a single Ginkgo CLI that is currently
  unsupported.
  
[1732290406] Webhook Suite - 3/3 specs ••• SUCCESS! 8.971230559s PASS
[1732290406] Cnihelper Suite - 1/1 specs • SUCCESS! 984.902µs PASS
...
```

bn222: `make gingko`

thom311: do you mean, to add a make target for calling `go get ginko`?

bn222: I mean that we already have that target.

thom311: I think that is not correct.

bn222: what's wrong with that?

thom311: there is no make target `make gingko` in dpu-operator `main` branch.

bn222: https://github.com/openshift/dpu-operator/blob/main/Makefile#L342

internal/controller/suite_test.go
```diff
@@ -33,7 +32,6 @@ import (
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
 var cfg *rest.Config
-var testEnv *envtest.Environment
```
bn222, line 36: Why do we need to expose it?

bn222:
/hold

Not holding because of contents, but rather because e2e ci seems not to be checking this anymore.

internal/controller/suite_test.go
thom311: I dropped the patch.

api/v1/dpuoperatorconfig_webhook_test.go
```diff
@@ -18,21 +18,95 @@ package v1
...
 
-		})
+			config := &DpuOperatorConfig{}
+
+			key := types.NamespacedName{
+				Name:      "dpu-operator-config",
+				Namespace: "openshift-dpu-operator",
+			}
+
+			err = k8sClient.Get(ctx, key, config)
```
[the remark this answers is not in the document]
thom311: I don't understand this question. The test runs against the test server that the test starts.

internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -231,9 +231,25 @@ var _ = Describe("Main Controller", Ordered, func() {
...
 				Expect(errors.IsNotFound(err)).To(BeTrue())
 			})
+			It("only DpuOperatorConfig \"openshift-dpu-operator/dpu-operator-config\" is allowed", func() {
+				ns := dpuOperatorNameSpace()
+				client := mgr.GetClient()
+
+				/* The CR already exists. Adding it again will fail. */
+				cr := dpuOperatorCR(testDpuOperatorConfigName, "host", ns)
+				err := client.Create(context.Background(), cr)
+				Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
thom311: I removed asserting for this.

internal/controller/dpuoperatorconfig_controller_test.go
```diff
@@ -231,9 +231,25 @@ var _ = Describe("Main Controller", Ordered, func() {
...
+
+				cr2 := dpuOperatorCR("foo", "host", ns)
+				err2 := client.Create(context.Background(), cr2)
+				/* The validating webhook does not run in this setup. Otherwise we would
+				 * expect creating this CR to fail. */
```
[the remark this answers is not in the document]
thom311: I reworded the commit message and this code comment.

As commit message says, that I was unable to make it work. The commit message says, that work is not currently planned, but could be done in the future (if somebody figures out how to make it work).

Probably it should be something like the following (which doesn't work):

```diff
From 1cce9259b8db7c250ae422f004313ca424f914a8 Mon Sep 17 00:00:00 2001
From: Thomas Haller <thaller@redhat.com>
Date: Mon, 25 Nov 2024 22:30:22 +0100
Subject: [PATCH 1/2] testutils: fix global variable for TestEnv

We will actually need this global variable, that gets initialized
during BeforeSuite(). Add it as "testutils.TestEnv".
---
 internal/controller/suite_test.go  | 2 --
 internal/testutils/test_cluster.go | 6 ++++--
 2 files changed, 4 insertions(+), 4 deletions(-)

diff --git a/internal/controller/suite_test.go b/internal/controller/suite_test.go
index 41ef6740831f..1a1e1b87d90e 100644
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -23,7 +23,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	"k8s.io/client-go/rest"
-	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	//+kubebuilder:scaffold:imports
@@ -33,7 +32,6 @@ import (
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
 var cfg *rest.Config
-var testEnv *envtest.Environment
 
 func TestControllers(t *testing.T) {
 	RegisterFailHandler(Fail)
diff --git a/internal/testutils/test_cluster.go b/internal/testutils/test_cluster.go
index 84b7e6dd746a..5739b121d232 100644
--- a/internal/testutils/test_cluster.go
+++ b/internal/testutils/test_cluster.go
@@ -37,6 +37,8 @@ var (
 	setupLog                  = ctrl.Log.WithName("setup")
 )
 
+var TestEnv *envtest.Environment
+
 func relativeToAbs(path string) string {
 	_, file, _, _ := runtime.Caller(0)
 	file, err := filepath.Abs(file)
@@ -48,7 +50,7 @@ func bootstrapTestEnv(restConfig *rest.Config) {
 	var err error
 	trueVal := true
 	By("bootstrapping test environment")
-	testEnv := &envtest.Environment{
+	TestEnv = &envtest.Environment{
 		CRDDirectoryPaths: []string{
 			relativeToAbs("../../config/crd/bases"),
 			relativeToAbs("../../test/crd"),
@@ -58,7 +60,7 @@ func bootstrapTestEnv(restConfig *rest.Config) {
 		Config:                restConfig,
 	}
 	By("starting the test env")
-	cfg, err := testEnv.Start()
+	cfg, err := TestEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 }

base-commit: cccb01b501f633015d7e48d1afb2fb66c60b66cd
-- 
2.47.1


From 62afdecc264dccbf2a3fc8836c4ee1940d20d89a Mon Sep 17 00:00:00 2001
From: Thomas Haller <thaller@redhat.com>
Date: Mon, 2 Dec 2024 17:53:26 +0100
Subject: [PATCH 2/2] controller/test: enable controller-test for new webhook

This doesn't actually work.
---
 .../dpuoperatorconfig_controller_test.go      | 25 ++++++++++++++++++-
 internal/testutils/test_cluster.go            |  3 +++
 2 files changed, 27 insertions(+), 1 deletion(-)

diff --git a/internal/controller/dpuoperatorconfig_controller_test.go b/internal/controller/dpuoperatorconfig_controller_test.go
index 6c3a834899f4..497fc3c1b714 100755
--- a/internal/controller/dpuoperatorconfig_controller_test.go
+++ b/internal/controller/dpuoperatorconfig_controller_test.go
@@ -2,8 +2,12 @@ package controller
 
 import (
 	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
 	"os"
 	"sync"
+	"time"
 
 	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
@@ -120,16 +124,24 @@ func startDPUControllerManager(ctx context.Context, client *rest.Config, wg *syn
 	utilruntime.Must(configv1.AddToScheme(scheme))
 	utilruntime.Must(netattdefv1.AddToScheme(scheme))
 
+	webhookInstallOptions := &testutils.TestEnv.WebhookInstallOptions
 	mgr, err := ctrl.NewManager(client, ctrl.Options{
 		Scheme: scheme,
 		Metrics: server.Options{
 			BindAddress: ":18001",
 		},
-		WebhookServer:    webhook.NewServer(webhook.Options{Port: 9443}),
+		WebhookServer: webhook.NewServer(webhook.Options{
+			Host:    webhookInstallOptions.LocalServingHost,
+			Port:    webhookInstallOptions.LocalServingPort,
+			CertDir: webhookInstallOptions.LocalServingCertDir,
+		}),
 		LeaderElectionID: "1e46962d.openshift.io",
 	})
 	Expect(err).NotTo(HaveOccurred())
 
+	err = (&configv1.DpuOperatorConfig{}).SetupWebhookWithManager(mgr)
+	Expect(err).NotTo(HaveOccurred())
+
 	b := NewDpuOperatorConfigReconciler(mgr.GetClient(), mgr.GetScheme(), "mock-image", plugin.CreateVspImagesMap(false, setupLog))
 	err = b.SetupWithManager(mgr)
 	Expect(err).NotTo(HaveOccurred())
@@ -143,6 +155,17 @@ func startDPUControllerManager(ctx context.Context, client *rest.Config, wg *syn
 	}()
 	<-mgr.Elected()
 
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := fmt.Sprintf("%s:%d", webhookInstallOptions.LocalServingHost, webhookInstallOptions.LocalServingPort)
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		return conn.Close()
+	}).Should(Succeed())
+
 	return mgr
 }
 
diff --git a/internal/testutils/test_cluster.go b/internal/testutils/test_cluster.go
index 5739b121d232..eea46b5e9db3 100644
--- a/internal/testutils/test_cluster.go
+++ b/internal/testutils/test_cluster.go
@@ -58,6 +58,9 @@ func bootstrapTestEnv(restConfig *rest.Config) {
 		ErrorIfCRDPathMissing: true,
 		UseExistingCluster:    &trueVal,
 		Config:                restConfig,
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			Paths: []string{relativeToAbs("../../config/webhook")},
+		},
 	}
 	By("starting the test env")
 	cfg, err := TestEnv.Start()
-- 
2.47.1

```

api/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+	scheme := apimachineryruntime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = admissionv1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	//+kubebuilder:scaffold:scheme
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
```
bn222, line 98: Run against kind cluster instead.

api/v1/dpuoperatorconfig_webhook_test.go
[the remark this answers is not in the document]
bn222: https://github.com/openshift/dpu-operator/pull/213#discussion_r1896121836

api/v1/webhook_suite_test.go
thom311: how to do that?

bn222:
cancelled ci because we need to get vsp as lib in first

internal/controller/dpuoperatorconfig_controller_test.go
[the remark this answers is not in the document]
thom311: @bn222 my point is: I tried for a long time (and failed) to add the validating webhook to `dpuoperatorconfig_controller_test.go`.

If you think that is necessary, then I'd need some pointers (possibly building on the patch ^^).

But it shouldn't be necessary, because:
- there is a working unit test that has the validating webhook running. This one does test the webhook.
- a validating webhook can only reject a CR. Having that in `dpuoperatorconfig_controller_test.go` only makes a difference, when trying to configure invalid CRs. Which we can do on the other unit test.
- extending the tests can be done later at any time.

bn222:
@vrindle you're in an excellent position to review this in detail since you've approached this differently.

bn222:
If we can't get his working in kind, we need at least e2e test. Can you add tests against a cda cluster?

Definitely remove the piece of code that doesn't make sense:

https://github.com/openshift/dpu-operator/pull/213/files#diff-24489aa02e68225d41d99794e841fe30ffb62033024bd3dd2e66fa9cc364035eR242-R244

thom311:
> If we can't get this working in kind, we need at least e2e test. Can you add tests against a cda cluster?

done

e2e_test/e2e_test.go
```diff
@@ -69,6 +70,74 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+		var config configv1.DpuOperatorConfig
+		var dpuOperatorConfig1 configv1.DpuOperatorConfig
+
+		crNamespace := "openshift-dpu-operator"
+		crName := "dpu-operator-config"
+
+		dpuOperatorConfig0 := configv1.DpuOperatorConfig{}
+
+		// If this failed, do you have a DpuOperatorConfig in your cluster?
+		// Create it with `oc create -f ./examples/dpu.yaml`.
```
bn222, line 84: Create it instead of telling the user to create it.

thom311:
> Definitely remove the piece of code that doesn't make sense:
>
> https://github.com/openshift/dpu-operator/pull/213/files#diff-24489aa02e68225d41d99794e841fe30ffb62033024bd3dd2e66fa9cc364035eR242-R244

Could you please clarify what you are linking at? I dropped the commit `controller/test: adjust controller-test for new webhook`, if that is what you meant. Obviously, I didn't think that "[that] doesn't make sense".

e2e_test/e2e_test.go
```diff
@@ -69,6 +70,74 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+		Eventually(func() error {
+			return dpuSideClient.Get(context.TODO(), types.NamespacedName{Namespace: crNamespace, Name: crName}, &dpuOperatorConfig0)
+		}, testutils.TestAPITimeout, testutils.TestRetryInterval).Should(Succeed())
+
+		// We delete the DpuOperatorConfig and restore it at the end. This is
+		// dangerous, if the test fails, we end up deleting the
+		// DpuOperatorConfig on the cluster and the user needs to manually fix
+		// it.
```
bn222, line 92: Write the code so that it isn't dangerous.

e2e_test/e2e_test.go
```diff
@@ -69,6 +70,74 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+		config = configv1.DpuOperatorConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: crNamespace,
+				Name:      crName,
+			},
+			Spec: configv1.DpuOperatorConfigSpec{
+				Mode:     "invalidmode",
+				LogLevel: 2,
+			},
+		}
```
bn222, line 122: Helper function to create a DpuOperatorConfig CR? You're using it twice here.

e2e_test/e2e_test.go
```diff
@@ -69,6 +70,74 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+				Name:      crName,
+			},
+			Spec: dpuOperatorConfig0.Spec,
+		}
+		err = dpuSideClient.Create(context.TODO(), &config)
+		Expect(err).NotTo(HaveOccurred())
+		dpuOperatorConfig1 = configv1.DpuOperatorConfig{}
+		Eventually(func() error {
+			return dpuSideClient.Get(context.TODO(), types.NamespacedName{Namespace: crNamespace, Name: crName}, &dpuOperatorConfig1)
+		}, testutils.TestAPITimeout, testutils.TestRetryInterval).Should(Succeed())
```
bn222, line 138: If we have an `Eventually` here, why don't we have it in between the calls to `Create`?

bn222:
> > Definitely remove the piece of code that doesn't make sense:
> > https://github.com/openshift/dpu-operator/pull/213/files#diff-24489aa02e68225d41d99794e841fe30ffb62033024bd3dd2e66fa9cc364035eR242-R244
> 
> Could you please clarify what you are linking at? I dropped the commit `controller/test: adjust controller-test for new webhook`, if that is what you meant. Obviously, I didn't think that "[that] doesn't make sense".

You were adding a test that would test if something was in an improper state. I recall you couldn't get the webhook working against a test cluster and had to write `Expect(..).NotTo` instead of `Expect(..).To` just to make the test pass. Is that the code you removed?

e2e_test/e2e_test.go
thom311: done.

e2e_test/e2e_test.go
thom311: I changed that.

The "Should create a pod when creating an SFC" has exactly the same problem, that if the test fails it leaves the sfc CR in the cluster, and a subsequent test run fails (without a code comment that would be helpful to identify what's wrong).

Anyway, I change the new test. It will now work regardless whether a DpuOperatorConfig already exists, it will temporarily delete it, and will restore the previous state afterwards.

bn222:
> I don't see the value of dropping tests that were already written (and pass).

More tests is not always better. Having tests in place makes a statement about how the code should work. Two relevant aspects are important here: 

- First, if the behavior is wrong, and you add a test just for the sake of adding a test, then the test will test that the behavior is wrong. In our case here, if a kind cluster cannot deal with webhooks, then the test against it will need to enforce the wrong behavior
- If you test a helper function, then now that helper function becomes an API. Helper functions are details. This point will only come across if you ever have to make changes in a project somewhere in a helper function and you end up spending hours to fix all kinds of tests that fail because the internals change without have any material effect on how the software as a whole works.

api/v1/dpuoperatorconfig_webhook_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	config.SetName("dpu-operator-config2")
+	_, err = config.validateDpuOperatorConfig()
+	Expect(err).To(HaveOccurred())
+}
+
+var _ = Describe("DpuOperatorConfig Webhook", func() {
+
+	Context("When creating DpuOperatorConfig under Validating Webhook", func() {
+		It("Validation should happen", func() {
```
bn222, line 48: This lumps together a bunch of tests. If it fails, the BDD framework won't provide very useful errors. Let's split it up and be more precise than "Validation should happen".

api/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "bin", "k8s",
+			fmt.Sprintf("1.29.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
+
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			Paths: []string{filepath.Join("..", "..", "config", "webhook")},
+		},
+	}
```
bn222, line 81: We've removed the use of testEnv and have an API to access either a KindCluster or a CdaCluster. The reason is that testEnv has a few gaps that make testing all features hard. If we start adding back testEnv, it will communicate to the next developer that this is useful to use throughout. They will lose quite some time if something doesn't work as expected. Either use KindCluster or CdaCluster, but not envtest.Environment directly.

api/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	var err error
+	// cfg is defined in this file globally.
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
```
bn222, line 87: This will automatically be cleaned up if you take into account the above comment

api/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	scheme := apimachineryruntime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
```
bn222, line 91: Same here, we use a scheme elsewhere that's now managed in one place.

api/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	err = admissionv1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
```
bn222, line 95: Same here, no need to copy `AddToScheme` all over the code base.

api/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+	//+kubebuilder:scaffold:scheme
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
```
bn222, line 100: Again, if you start using `KindCluster` (preferably) or `CdaCluster` (if kind has limitations), then this will be cleaned up too.

e2e_test/e2e_test.go
bn222: Are you preparing a follow up PR immediately after this?

e2e_test/e2e_test.go
```diff
@@ -67,6 +71,86 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			config = configv1.DpuOperatorConfig{}
+			Eventually(func() error {
+				return dpuSideClient.Get(context.TODO(), types.NamespacedName{Namespace: crNamespace, Name: crName}, &config)
+			}, testutils.TestAPITimeout, testutils.TestRetryInterval).Should(Succeed())
+		}
+
+	})
+
+	g.It("Should run Webhook to validate DpuOperatorConfig", func() {
```
bn222, line 94: Can we move this into another file, ie `e2e_test/...`

e2e_test/e2e_test.go
```diff
@@ -67,6 +71,86 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+			},
+			Spec: configv1.DpuOperatorConfigSpec{
+				Mode:     "host",
+				LogLevel: 2,
+			},
+		}
+		err = dpuSideClient.Create(context.TODO(), &config)
+		Expect(err).To(HaveOccurred())
+
+		config = configv1.DpuOperatorConfig{
```
bn222, line 123: Don't reuse the same var name.

e2e_test/e2e_test.go
```diff
@@ -67,6 +71,86 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+			},
+			Spec: configv1.DpuOperatorConfigSpec{
+				Mode:     "invalidmode",
+				LogLevel: 2,
+			},
+		}
+		err = dpuSideClient.Create(context.TODO(), &config)
+		Expect(err).To(HaveOccurred())
+
+		config = configv1.DpuOperatorConfig{
```
bn222, line 136: Same here, don't reuse the same var name. Use names like `goodConfig`, `badConfig`, `invalidConfig`

bn222 (commented):
Another round of reviews. Can you also squash some of the patches so that changes are grouped together logically?

api/v1/dpuoperatorconfig_webhook_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+			config := &DpuOperatorConfig{}
+
+			key := types.NamespacedName{
+				Name:      "dpu-operator-config",
+				Namespace: "openshift-dpu-operator",
+			}
+
+			err = k8sClient.Get(ctx, key, config)
+			Expect(err).To(HaveOccurred())
```
bn222, line 61: It's not clear to me what we're testing here

thom311: Because when the test starts, no such DpuOperatorConfig exists yet.

it asserts, that it does not yet exist.


(as always, every assert needs to be side-effect free and must not fail. So no assert is ever truly necessary. Having asserts in a test is still useful, even if some asserts "obviously" hold -- they all hold, to some degree of "obviously").

bn222: Can we move it to a block that checks if the context is as expected?
`When started` and `No DpuOperatorConfig exists`

e2e_test/e2e_test.go
thom311: that is not planned (there is no Jira issue for that).

api/v1/webhook_suite_test.go
thom311: this file is generated by operator-sdk (I didn't touch it). It is their suggested way for writing unit tests for the webhook.

Since we use(d) operator-sdk, we should not interfere with the generated code or counter their suggested ways of doing it.

This does not mean, there cannot be other tests beside that (as there are, with KindCluster/CdaCluster). There is no conflict here, except that this is yet another test. This test will obviously be more unit-like, self-contained, contrary to the more integration-like tests in CdaCluster. Both are useful to have.

Also, the main maintainance burden is with CdaCluster, which uses an external cluster, and can easily break. E.g. to break `"Should create a pod when creating an SFC"` just have a left-over SFC. This is what makes those tests hard(er) to maintain -- but for good reasons, as their tests are still valuable. There is less reason why this (more self-contained) unit test would fail and the maintenance burden is negligible.

Feel free to delete tests afterwards. Tests can always be deleted (too), though, I disagree with doing that.

api/v1/webhook_suite_test.go
thom311: This code was generated by commit "webhook: generate webhook using operator-sdk".

That commit is *entirely* generated, I didn't manually write a single line. The command to (re)generate the commit is in the commit message, so the main angel to review this generated patch is by reviewing the commands that generated it.

Beyond that, you can of course review what was generated too. And, if you want, there can be a future commit that deletes the generated test code (though I disagree with doing that).

api/v1/webhook_suite_test.go
thom311: https://github.com/openshift/dpu-operator/pull/213#discussion_r1961527131

api/v1/webhook_suite_test.go
thom311: https://github.com/openshift/dpu-operator/pull/213#discussion_r1961527131

api/v1/webhook_suite_test.go
thom311: https://github.com/openshift/dpu-operator/pull/213#discussion_r1961527131

api/v1/dpuoperatorconfig_webhook_test.go
thom311: done.

e2e_test/e2e_test.go
thom311: the file is called `e2e_test/e2e_test.go`. So presumably it is a place for running various (related) tests.

The pre-existing test is about SFC, but the file is not called `e2e_test/e2e_test_sfc.go` either.

Before this PR, `e2e_test/e2e_test.go` contains one test, and (estimated) 45% of it are just boilerplate concerned with import of modules and other setup. By adding a new file, that file will contain about same boilerplate.

I'd rather leave it here. It's a fine place IMO.

api/v1/dpuoperatorconfig_webhook_test.go
thom311: added a new `g.It()` "block".

api/v1/webhook_suite_test.go
bn222: Can you elaborate on why you disagree with doing that?

api/v1/webhook_suite_test.go
bn222: > we should not interfere with the generated code

Can you elaborate on this?

e2e_test/e2e_test.go
bn222: I like the idea to rename the current file to `e2e_test/e2e_test_sfc.go` first, and add another one for your tests. Common parts can be part of the suite wide functions to avoid duplication.

api/v1/dpuoperatorconfig_webhook_test.go
bn222: Unless I'm missing something, I don't see the new `It` block.

bn222:
@SalDaniele @wizhaoredhat, could you please review this as well and add your comments? The current open points are:
1. Should we re-add the `envtest`-related code after we intentionally remove it elsewhere and wrapped it in our test cluster framework to avoid future contributors wasting time when they unexpectedly encounter the limitations of `envtest`?
2. Split up the test files so that the filename indicates the group of tests being run. The main goal is to avoid test files that become too long.
3. Create a follow-up work item to add a helper function to create a `DpuOperatorConfig` CR, or include that small change in this PR.

SalDaniele:
> @SalDaniele @wizhaoredhat, could you please review this as well and add your comments? The current open points are:
> 
> 1. Should we re-add the `envtest`-related code after we intentionally remove it elsewhere and wrapped it in our test cluster framework to avoid future contributors wasting time when they unexpectedly encounter the limitations of `envtest`?

Added comments in review

> 2. Split up the test files so that the filename indicates the group of tests being run. The main goal is to avoid test files that become too long.

I think this refers to the e2e-test being added? I am fine with merging this PR as is, as we add more e2e tests in the future, splitting this up by file rather than context makes a lot of sense to me. Maybe it is best to just start this with this PR (update existing e2e-test to explicit SFC-e2e-test, add webhook-e2e-test)

> 3. Create a follow-up work item to add a helper function to create a `DpuOperatorConfig` CR, or include that small change in this PR.

Sounds good, either way. If this PR works now, I am in favor of merging ASAP so we can avoid the pain of constant rebasing.

api/v1/webhook_suite_test.go
bn222: Ok, sounds good to push another pr to remove the envtest code and replace it with kind. 

@wizhaoredhat I'd like to get your input here as well since you went through the debugging session when we tried to use envtest

wizhaoredhat:
> @SalDaniele @wizhaoredhat, could you please review this as well and add your comments? The current open points are:
> 
> 1. Should we re-add the `envtest`-related code after we intentionally remove it elsewhere and wrapped it in our test cluster framework to avoid future contributors wasting time when they unexpectedly encounter the limitations of `envtest`?
> 2. Split up the test files so that the filename indicates the group of tests being run. The main goal is to avoid test files that become too long.
> 3. Create a follow-up work item to add a helper function to create a `DpuOperatorConfig` CR, or include that small change in this PR.

1. The limitations of envtest comes from "cleaning up" and reusing the same envtest for other tests. 
If envtest is created independently for each individual test then it should work fine. The effect of other tests should not leak into the test that is being run. Contributors on the contrary  are used to envtest. Most of the k8s projects I have seen use this and also operator-sdk generates code scaffolding that uses it. This is what Thomas did, he used operator-sdk (widely used in k8s).

api/v1/webhook_suite_test.go
wizhaoredhat: This was the issue with envtest: https://book.kubebuilder.io/reference/envtest.html#:~:text=EnvTest%20does%20not%20support%20namespace%20deletion.

bn222:
@wizhaoredhat we're still using envtest, but we test against a kind cluster: https://github.com/openshift/dpu-operator/blob/7d42a16e4f09430a1ca3ab4302d2db7a47578ee6/internal/testutils/kindcluster.go#L100 . 

If we want an EnvTestCluster struct, we should add it so that we can also test against a kind cluster and a real cluster easily. 

The benefit of calling the TestCluster api instead of envtest directly is that we can change the cluster on which we test: kind, real, envtest. The code in this pr is limited to envtest. The DPU operator interfaces with real hardware. In that, it's different from many other projects.

bn222:
`make vendor-check` is failing. We should also add `make bundle-check` since that would fail too in the current version

bn222:
https://github.com/openshift/dpu-operator/blob/88cc5cfddb4964cb4bd3599dc5abbc01c34a1e5c/api/v1/dpuoperatorconfig_webhook_test.go#L81
Race condition when deleting. Need "eventually it should be in a particular state" check.

e2e_test/e2e_test.go
```diff
@@ -70,6 +78,150 @@ var _ = g.Describe("Dpu side", g.Ordered, func() {
...
+				 * has a DpuOperatorConfig. This test wants to create different DpuOperatorConfig
+				 * to check whether the validating webhook is running. For that, it first needs
+				 * to delete (and later restore, see "restoreDpuOperatorConfigInAfterAll") the
+				 * DpuOperatorConfig.
+				 *
+				 * However, the followup test for SFC has a problem with that, and will fail
+				 * when recreating the DpuOperatorConfig.
+				 *
+				 * As temporary workaround, skip.
+				 */
```
SalDaniele, line 99: LGTM for now...I think the issue is the Intel VSP fails if initialized twice. This is a bug that needs to be addressed by Intel before we can add this back in

SalDaniele:
/test make-e2e-test

Hit an unrelated flake in CDA

SalDaniele:
/test make-e2e-test

idpf flake

SalDaniele:
/test make-e2e-test
```

Openshift installation failed with a new error:

[core@ocpcluster-master-2 ~]$ journalctl -b -f -u release-image.service -u bootkube.service
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Checking if api.ocpcluster.redhat.com of type API_URL reachable
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Unable to reach API_URL's https endpoint at https://api.ocpcluster.redhat.com:6443/readyz
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Unable to validate. https://api.ocpcluster.redhat.com:6443/readyz is currently unreachable.
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Check if API-Int URL is reachable during bootstrap
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Checking if api-int.ocpcluster.redhat.com of type API_INT_URL reachable
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Unable to reach API_INT_URL's https endpoint at https://api-int.ocpcluster.redhat.com:6443/readyz
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: Unable to validate. https://api-int.ocpcluster.redhat.com:6443/readyz is currently unreachable.
Feb 27 23:38:27 ocpcluster-master-2 bootkube.sh[3784]: bootkube.service complete
Feb 27 23:38:27 ocpcluster-master-2 systemd[1]: bootkube.service: Deactivated successfully.
Feb 27 23:38:27 ocpcluster-master-2 systemd[1]: bootkube.service: Consumed 32.631s CPU time.
```

PR #214, opened 2024-11-21
Introduce afero in allocator

bn222:


SalDaniele:
I think there are still some utilities in the CNI that require the hardware / reference "os" if the goal here is remove this entirely. i.e. paths:https://github.com/openshift/dpu-operator/blob/main/dpu-cni/pkgs/sriovutils/sriovutils.go.

But this makes sense to me as a good step

bn222:
> I think there are still some utilities in the CNI that require the hardware / reference "os" if the goal here is remove this entirely. i.e. https://github.com/openshift/dpu-operator/blob/main/dpu-cni/pkgs/sriovutils/sriovutils.go.
> 
> But this makes sense to me as a good step

This is precisely what the next step is about, but the current PR is a piecemeal step.

bn222:
Stripped this down to the bare minimum. @SalDaniele ptal

dpu-cni/pkgs/sriovutils/pci_allocator.go
```diff
@@ -2,91 +2,57 @@ package sriovutils
...
-	if err != nil {
-		if os.IsNotExist(err) {
-			return false, nil
-		}
-
-		return false, fmt.Errorf("failed to check for pci address file for %s: %v", path, err)
-	}
+func namespaceExists(namespace string) bool {
+	networkNamespace, err := ns.GetNS(namespace)
+	ret := err != nil
```
SalDaniele, line 36: Shouldn't we check whether err != nil here?

Otherwise GetNS could do something unexpected, and networkNamespace.Close() might cause a  nil pointer dereference.

Maybe:

```
func namespaceExists(namespace string) bool {
	networkNamespace, err := ns.GetNS(namespace)
	if err != nil {
		return false
	}
	defer networkNamespace.Close()
	return true
}
```

bn222: good catch

bn222: fixed.

wizhaoredhat:
Please add a more meaningful commit message detailing why this change is needed, what does it improve, etc...

PR #224, opened 2024-12-13
[th/buildah-no-authfile] build: don't use `--authfile /root/config.json` for buildah commands

thom311:
"/root/config.json" is only accessible to root, which means it makes the build not usable to regular users. That is bad.

Instead, if the build will require a login, then require the caller to first issue a suitable `podman login registry.ci.openshift.org` or configure podman in a similar way, so that a subsequent pull works.  If the user prefers to use files, they still can put the auth file to the well known place, but that place is "/run/user/$UID/containers/auth.json" or "~/.config/containers/auth.json", not "/root/config.json".

That is essentially the same as before: the caller must prepare the system in a way so that pull is gonna work. Except, now they have all the options that podman (e.g. `podman login`) offers, instead of hard coding some file at "/root/config.json".

You can migrate those machines with:

    #BASEDIR=/run/user/0
    BASEDIR=~/.config
    mkdir -p $BASEDIR/containers
    chmod 700 $BASEDIR/containers
    cp /root/config.json $BASEDIR/containers/auth.json
    chmod 600 $BASEDIR/containers/auth.json

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,16 +267,16 @@ local-buildx: prepare-multi-arch ## Build all container images necessary to run 
 	mkdir -p $(GO_CONTAINER_CACHE)
 	buildah manifest rm $(DPU_OPERATOR_IMAGE)-manifest || true
 	buildah manifest create $(DPU_OPERATOR_IMAGE)-manifest
-	buildah build --layers --manifest $(DPU_OPERATOR_IMAGE)-manifest --platform linux/amd64,linux/arm64 -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
+	buildah build $(LOCAL_BUILDX_ARGS) --layers --manifest $(DPU_OPERATOR_IMAGE)-manifest --platform linux/amd64,linux/arm64 -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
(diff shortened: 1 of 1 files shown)


thom311:
If this were merged, then the systems that previously successfully ran `make build-localx` (or use that via CDA's `dpu_operator_{host,dpu}` postconfig), need to be adjusted, so that the authentication file is found. As described in the commit message. That may affect CI systems.

bn222:
Thanks. Can we make it better yet logging in automatically while building from CDA (if not yet logged in) so that we don't have to adjust our CI servers for the change of path manually?

thom311:
> Thanks. Can we make it better by logging in automatically while building from CDA (if not yet logged in) so that we don't have to adjust our CI servers for the change of path manually?

I don't see how exactly, for one, we wouldn't want to hard code secrets in CDA or our CI.

Also, it's not necessary. Latest dpu-operator sets `{builder,base}_image` to some other repo. That doesn't require authentication. On which system is it necessary? The current approach is that the admin of the system pre-sets up the file. We can go through the handful of systems that are affected (are there any?) and migrate manually, as describe in the commit message.

bn222:
rebase and rename `DPU_OPERATOR_LOCAL_BUILDX_ARGS` to `LOCAL_BUILDX_ARGS`

thom311:
> rebase and rename `DPU_OPERATOR_LOCAL_BUILDX_ARGS` to `LOCAL_BUILDX_ARGS`

done.

PR #225, opened 2024-12-13
hack: Be explicit with image rebuild behavior

SalDaniele:
Ensure that we only rebuild the images once per CI run. Include a bump to CDA to ensure the correct images are available when testing.


SalDaniele:
@vrindle @wizhaoredhat PTAL, need lgtm

PR #227, opened 2024-12-16
ipu-vsp: Add Peer2Peer P4 rules independant...

SalDaniele:
Add Peer2Peer P4 rules independant of order of deployment
Fixing package installs in p4sdk to ubi repo from rhel-9*


sudhar-krishnakumar:
> @sudhar-krishnakumar PTAL. In particular, verify the daemonset is as expected given the vsp change.
> 
> Also if possible please link any associated jira tied to this change (could not find any such ticket), so we can move to closed pending this passing.

@SalDaniele We had an internal JIRA on our side, looks like we dont have any such common JIRA...so not able to put a link here.

sudhar-krishnakumar (approved):
@SalDaniele , Looks good. Thanks Sal.

PR #229, opened 2024-12-17
[IIC-425] Remove race condition between pod/vsp

SalDaniele:
This commit introduces intel/ipu-opi-plugins@81c6a0e to remove the race condition between the vsp and p4 pod initializing.

We also remove the corresponding workaround in the automation script to wait for the p4 to finish initializing before proceeding.


SalDaniele:
@saynb PTAL

saynb (approved):

PR #230, opened 2024-12-18
[IIC-427] Remove need to run ipu_port1 manually

SalDaniele:
We should no longer need to hack a script to ensure primary networking is working at all times. This should be handled by the VSP now.

### DIFF
--- a/hack/cluster-configs/config-dpu-host.yaml
+++ b/hack/cluster-configs/config-dpu-host.yaml
@@ -9,7 +9,7 @@ clusters:
     - name: "dpu_operator_host"
       dpu_operator_path: "../../"
       rebuild_dpu_operators_images: "false"
-      ipu_plugin_sha: "81c6a0e9fe8cefee4cdac88a6a20bbc6b81e0df3"
+      ipu_plugin_sha: "f01e94e22c07a54d3d4f26dca28e89b65cb3678f"
       builder_image: "wsfd-advnetlab-amp04.anl.eng.bos2.dc.redhat.com:5000/ocp/builder:rhel-9-golang-1.22-openshift-4.18"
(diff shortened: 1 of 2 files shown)


SalDaniele:
@sudhar-krishnakumar PTAL

SalDaniele:
Hi @sudhar-krishnakumar this commit is failing. If we remove the CDA workarounds [1] to ensure port connectivity on ACC reboot, the ACC does not have connectivity after re-installing RHEL.

I believe this is a chicken and egg issue. We do not get the post install script installed by the vsp, because we do not get to the point that the vsp is run.

Is the expectation that we remove the fxp-networking package from all CI / Dev servers manually, then re-deploy with this new vsp? At this point, we will be able to assume that the ACC will always have connectivity across all states:

i.e.
- I deploy vsp for first time, vsp triggers IMC reboot and installs fxp-linux package, ACC should come up good (note we cannot test this programmatically in the CI because of [2], which means on ACC reboot the solution does not come up automatically.)

- I was previously on fxp-linux, and I want to re-install RHEL on the ACC

- I cold boot the host

- I reboot the IMC

Please advise.

[1] https://github.com/bn222/cluster-deployment-automation/commit/44a6c1cd041662ecd88661183ae85296f5075c7e
[2] https://issues.redhat.com/browse/IIC-465

SalDaniele:
Advice from @sudhar-krishnakumar  is to leave the workaround to get the latest VSP changes running. If this passes, we can try removing the workaround in a subsequent PR.

We may need to manually add the post-init-script added by the new vsp to all CI / Dev servers in order for this to work.

bn222:
/lgtm

Essentially, we need to run this once on all servers. The proposal from @sudhar-krishnakumar is ok.

PR #231, opened 2024-12-18
Integrate Intel VSP as a lib

bn222:
Allow to build from another repo for VSP.


bn222:
@wizhaoredhat

bn222:
@vrindle when the CI is fixed and https://github.com/openshift/dpu-operator/pull/235 goes in, can you also give this one a retest?

cmd/intelvsp/p4runtime-2023.11.0-py3-none-any.whl
wizhaoredhat, line 1: What type of file is this? Is it a binary file?

bn222: Python dependency. It's called a "wheel" file. Intel knows they should remove this dependency soon since it's too opaque. It's a library that sits between the go code and the p4 runtime.

go.sum
```diff
@@ -1,3 +1,5 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
wizhaoredhat, line 1: This is a little bit strange to see. Any idea where this is pulled in?

Makefile
```diff
@@ -247,16 +247,16 @@ local-buildx: ## Build all container images necessary to run the whole operator
 	mkdir -p $(GO_CONTAINER_CACHE)
 	buildah manifest rm $(DPU_OPERATOR_IMAGE)-manifest || true
 	buildah manifest create $(DPU_OPERATOR_IMAGE)-manifest
-	buildah build --manifest $(DPU_OPERATOR_IMAGE)-manifest --platform linux/amd64,linux/arm64 -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
+	buildah build --layers --manifest $(DPU_OPERATOR_IMAGE)-manifest --platform linux/amd64,linux/arm64 -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.rhel -t $(DPU_OPERATOR_IMAGE)
```
wizhaoredhat, line 254: Maybe in the commit message mention what the improvement you see with this change to adding the --layers option.

Dockerfile.IntelVSP.rhel
```diff
@@ -0,0 +1,24 @@
...
+ENV P4_NAME=fxp-net_linux-networking
+RUN dnf install -y NetworkManager openvswitch3.4 iproute python3 python3-pip openssh-clients
+#By setting workdir, the directory is created automatically
+WORKDIR /opt/p4/p4-cp-nws/bin/
+
+COPY ./cmd/intelvsp/fxp-net_linux-networking.pkg /
+COPY ./cmd/intelvsp/p4rt-ctl /opt/p4/p4-cp-nws/bin/
+#TODO: Update to newer package, according to release.
+COPY ./cmd/intelvsp/p4runtime-2023.11.0-py3-none-any.whl /opt/p4/p4-cp-nws/bin/
+RUN python3 -m pip install --no-cache-dir /opt/p4/p4-cp-nws/bin/p4runtime-2023.11.0-py3-none-any.whl grpcio'=='1.59.3 protobuf'=='4.25.0 netaddr'=='0.9.0
```
wizhaoredhat, line 20: Is there plans to move away from python to golang? These untracked dependencies with Python may be a red flag for product security.

Makefile
```diff
@@ -242,8 +242,12 @@ local-build: ## Build all container images necessary to run the whole operator
 	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.mrvlVSP.rhel -t $(MARVELL_VSP_IMAGE)
 	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.IntelVSP.rhel -t $(INTEL_VSP_IMAGE)
 
+.PHONE: prepare-multi-arch
+prepare-multi-arch:
+	test -f /proc/sys/fs/binfmt_misc/qemu-aarch64 || sudo podman run --rm --privileged quay.io/bnemeth/multiarch-qemu-user-static --reset -p yes
```
wizhaoredhat, line 247: Is quay.io/bnemeth/multiarch-qemu-user-static going to be permanent?

Dockerfile.IntelVSP.rhel
bn222: This is an ongoing discussion with Intel https://issues.redhat.com/browse/IIC-442

Makefile
bn222: Yes

go.sum
bn222: It's being pulled in through Intel's VSP

Makefile
wizhaoredhat: We need some transparency to what this is if this step is going to be used to build production images.

wizhaoredhat: If it is only development, this should be good. But I think adding your containerfile in that is used to build multiarch-qemu-user-static in /hack would be good.

Makefile
bn222: added a commit message

Makefile
bn222: This is an external repo, I've added a commit message. If need be, we could add it as a submodule, and build and push the container as part of the build process. For now, I suggest we keep the commit message, and use the quay image I've prepared.

wizhaoredhat:
I don't see the new commit messages. Did you push it?

bn222:
I'm debugging it on a CI server now, there is something else wrong that I need to fix before I push again.

Makefile
```diff
@@ -242,8 +242,13 @@ local-build: ## Build all container images necessary to run the whole operator
...
```
wizhaoredhat, line 247: Looking at this a bit more, you are doing "prepare" step but it is not obvious where it is being used. Is there a specific ARM container you are trying to run on x86 build?

bn222: the prepare-multi-arch is a pre-req for the `local-buildx` target. This command needs to be executed to prepare buildah for the multi-arch build.

In the past we were lucky in that we didn't have `RUN` commands in the second part of our containers. Also, this will be a requirement for 4.19 going forward since they changed the builder layer image to be multi-arch too (that used to be x86 only, and we cross-compiled from x86 to arm in that layer _by chance_).

bn222:
/test make-e2e-test
(was missing `parallel` on our ci server)

PR #233, opened 2024-12-19
Add CBP(),DBP()  to Marvell VSP

alkama-hasan:
This Commits adds CreateBridgePort() and DeleteBridgePort() API Support
for Marvell VSP Additionally, it integrates the OVS with netdev or DPDK data plane
and  debug data plane functionality for better debugging  within Marvell VSP

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/debug-dp/debugdp.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/debug-dp/debugdp.go
@@ -1,0 +1,41 @@
+package DebugDP
+
+import (
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type DebugDP struct {
+	log logr.Logger
+}
+
+func NewDebugDP() *DebugDP {
+	return &DebugDP{
+		log: ctrl.Log.WithName("MarvellVSP:DebugDP"),
+	}
+}
+
+func (debugDP *DebugDP) AddPortToDataPlane(bridgeName string, portName string, vfPCIAddres string, isDPDK bool) error {
+	debugDP.log.Info("AddPortToBridge ", "bridgeName", bridgeName, "PortName", portName)
+	return nil
+}
+
+func (debugDP *DebugDP) DeletePortFromDataPlane(bridgeName string, portName string) error {
+	debugDP.log.Info("DeletePortFromBridge ", "bridgeName", bridgeName, "PortName", portName)
+	return nil
+
+}
+
+func (debugDP *DebugDP) InitDataPlane(bridgeName string) error {
+	debugDP.log.Info("Init Data plane", "bridgeName", bridgeName)
+	return nil
+}
+
+func (debugDP *DebugDP) ReadAllPortFromDataPlane(bridgeName string) (string, error) {
+	debugDP.log.Info("ReadAllPortFromBridge ", "bridgeName", bridgeName)
+	return "", nil
+}
+func (debugDP *DebugDP) DeleteDataplane(bridgeName string) error {
+	debugDP.log.Info("DeleteDataplane", "bridgeName", bridgeName)
+	return nil
+}
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -3,24 +3,29 @@ package main
 import (
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"net"
-	"os"
 	"os/exec"
-	"path/filepath"
+	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
-	ghw "github.com/jaypipes/ghw"
 	pb "github.com/openshift/dpu-operator/dpu-api/gen"
+	debugdp "github.com/openshift/dpu-operator/internal/daemon/vendor-specific-plugins/marvell/debug-dp"
+	mrvlutils "github.com/openshift/dpu-operator/internal/daemon/vendor-specific-plugins/marvell/mrvl-utils"
+	ovsdp "github.com/openshift/dpu-operator/internal/daemon/vendor-specific-plugins/marvell/ovs-dp"
 	"github.com/openshift/dpu-operator/internal/utils"
 	opi "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
 	"github.com/vishvananda/netlink"
+	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/protobuf/types/known/emptypb"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 const (
@@ -34,8 +39,20 @@ const (
 	NoOfPortPairs int    = 2
 	IPv6AddrDpu   string = "fe80::1"
 	IPv6AddrHost  string = "fe80::2"
+	DataPlaneType string = "debug"
+	NumPFs        int    = 1
+	PFID          int    = 0
+	isDPDK        bool   = false
 )
 
+// multiple dataplane can be added using mrvldp interface functions
+type mrvldp interface {
+	AddPortToDataPlane(bridgeName string, portName string, vfPCIAddres string, isDPDK bool) error
+	DeletePortFromDataPlane(bridgeName string, portName string) error
+	InitDataPlane(bridgeName string) error
+	ReadAllPortFromDataPlane(bridgeName string) (string, error)
+	DeleteDataplane(bridgeName string) error
+}
 type mrvlDeviceInfo struct {
 	nfInterfaceName string
 	dpInterfaceName string
@@ -58,6 +75,8 @@ type mrvlVspServer struct {
 	deviceStore   map[string]mrvlDeviceInfo
 	portType      string
 	noOfPortPairs int
+	bridgeName    string
+	mrvlDP        mrvldp
 }
 
 // createVethPair function to create a veth pair with the given index and InterfaceInfo
@@ -137,7 +156,7 @@ func (vsp *mrvlVspServer) ConfigureNetworkInterface() error {
 
 	switch vsp.portType {
 	case "veth":
-		klog.Infof("Creating %d veth pairs", vsp.noOfPortPairs)
+		klog.Infof("Creating Veth Pairs: %d", vsp.noOfPortPairs)
 		for i := 0; i < vsp.noOfPortPairs; i++ {
 			err := vsp.createVethPair(i)
 			if err != nil {
@@ -149,7 +168,7 @@ func (vsp *mrvlVspServer) ConfigureNetworkInterface() error {
 	case "hwlbk":
 		err := vsp.createHwLBK()
 		if err != nil {
-			klog.Errorf("Error occurred in creating HW Loopback: %v", err)
+			klog.Errorf("Error occurred in creating HW loopback: %v", err)
 			return err
 		}
 	default:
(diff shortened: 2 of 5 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -255,29 +360,54 @@ func (vsp *mrvlVspServer) GetDevices(ctx context.Context, in *pb.Empty) (*pb.Dev
...
+	}
+	pciAddress, err := mrvlutils.GetPCIByDeviceID(HostDeviceID)
+	if pciAddress == "" || err != nil {
+		return nil, errors.New("PCI Address not found")
+	}
+	vfcnt := in.VfCnt
+	if vfcnt < 0 {
+		return nil, errors.New("invalid VF Count")
+	}
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("echo %d > /sys/bus/pci/devices/0000:17:00.0/sriov_numvfs", vfcnt))
```
thom311, line 377: doesn't somebody need to take care that if sriov_numvfs is not zero, that it first gets reset to zero? Who does that or how is this supposed to work?

internal/daemon/vendor-specific-plugins/marvell/ovs-dp/ovsdp.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/go-logr/logr"
+	mrvlutils "github.com/openshift/dpu-operator/internal/daemon/vendor-specific-plugins/marvell/mrvl-utils"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	ovsDBPath  = "/var/run/openvswitch/db.sock"
+	ovsCLIPath = "/usr/local/bin/ovs-vsctl"
+	deviceID   = "a063"
```
thom311, line 15: I think constants should be CamelCase. Also because there is already `ovsDBPath` (in this spelling) used for something else.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -72,6 +90,9 @@ func (vsp *mrvlVspServer) createVethPair(index int) error {
 	if err := netlink.LinkAdd(vethLink); err != nil {
 		return err
 	}
+	if err := netlink.LinkSetUp(vethLink); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
alkama-hasan: No not necessary will update it

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -192,53 +213,137 @@ func (vsp *mrvlVspServer) Init(ctx context.Context, in *pb.InitRequest) (*pb.IpP
...
+
 	}
 	return &pb.IpPort{
 		Ip:   ipPort.Ip,
 		Port: ipPort.Port,
 	}, err
 }
 
+// getVFName function to get the VF Name of the given BridgePortName on DPU
+func (vsp *mrvlVspServer) getVFDetails(BridgePortName string) (string, string, error) {
```
wizhaoredhat, line 236: Would https://github.com/k8snetworkplumbingwg/sriovnet package be better?

internal/daemon/vendor-specific-plugins/marvell/ovs-dp/ovsdp.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ovs-vsctl command to create bridge
+func createBridge(bridgeName string) error {
+	cmd := exec.Command("ovs-vsctl", "--may-exist", "add-br", bridgeName, "--", "set", "bridge", bridgeName, "datapath_type=netdev")
+	return cmd.Run()
+}
+
+// InitDataPlane initializes the data path in this case it creates an ovs bridge
+func (ovsdp *OvsDP) InitDataPlane(bridgeName string) error {
+	ovsdp.log.Info("Initializing OVS-DPDK Data Plane")
```
wizhaoredhat, line 75: When taking with Sushil, he mentioned that only the non-performant netdev path has been upstreamed. Will we still use ovs-dpdk?

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -255,11 +360,45 @@ func (vsp *mrvlVspServer) GetDevices(ctx context.Context, in *pb.Empty) (*pb.Dev
...
+	if vfcnt < 0 {
+		return nil, errors.New("invalid VF Count")
+	}
+
+	// reset sriov_numvfs to 0 before setting to a number
+	resetCmd := exec.Command("sh", "-c", fmt.Sprintf("echo 0 > /sys/bus/pci/devices/%s/sriov_numvfs", pciAddress))
+	_, err = resetCmd.CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("failed to reset sriov_numvfs to 0: %v", err)
+	}
```
thom311, line 383: This looks better to me. Obviously, it needs to be tested whether it works well.

Resetting to zero is quite disruptive. Maybe the code should first read the file, and do nothing if it already contains the desired number.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -334,63 +474,14 @@ func (vsp *mrvlVspServer) hostIpPort() (pb.IpPort, error) {
...
-	ifname, err := GetPfName(pciAddress)
-	if err != nil {
-		return "", err
-	}
-	return ifname, nil
-}
-
 // enableIPV6LinkLocal function to enable the IPv6 Link Local Address on the given Interface Name
 // It will return the error
 func enableIPV6LinkLocal(interfaceName string) error {
```
thom311, line 487: Not directly related to this PR. However, `enableIPV6LinkLocal()` creates NetworkManager profiles. I don't think that is a good approach.

For one, that's only going to work with systems that have NetworkManager running. While that is the case for RHEL, optimally dpu-operator does not depend on that.

It's also slightly cumbersome to do (you would have to either bind-mount the D-Bus socket in the pod, or as you use `nsenter -t 1 -m -u -n -i -- nmcli ...` to be in the mount namespace of PID1. Additionally, you need suitable polkit permissions (or run as user root). That latter part is missing, so that we get :
```
# oc exec -ti pod/vsp-97tbs -- nsenter -t 1 -m -u -n -i -- nmcli connection add type ethernet
Error: Failed to add 'ethernet' connection: Insufficient privileges
command terminated with exit code 4
```

NetworkManager is IMO great, but not the right solution in this case because NetworkManager is about persistent configuration, when here we just want ad-hoc configure IP addresses.

I think it just needs something like:
```
ip link set $LINK addrgenmode eui64
ip link set $LINK down
ip link set $LINK up
```
or maybe use `ip token`

Let me try to send a patch for what I said here.

internal/daemon/vendor-specific-plugins/marvell/mrvl-utils/mrvlutils.go
```diff
@@ -0,0 +1,180 @@
...
+func GetNameByDeviceID(deviceID string) (string, error) {
+	targetVendorID := VendorID
+	targetDeviceID := deviceID
+	pci, err := ghw.PCI()
+	if err != nil {
+		return "", err
+	}
+	var pciAddress string
+	for _, device := range pci.Devices {
+		if device.Vendor != nil && device.Product != nil {
```
thom311, line 60: on the DPU, I get:
```
$ lspci -nn | grep a0f7
0002:01:00.1 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:00.2 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:00.3 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:00.4 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:00.5 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:00.6 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:00.7 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:01.0 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
0002:01:01.1 Ethernet controller [0200]: Cavium, Inc. Octeon Tx2 SDP Virtual Function [177d:a0f7] (rev 54)
```
so there are several PCI devices. How does the code know which one to take? I presume, it wants the PF?

alkama-hasan: yes it wants pf, the code takes first device of the array which is vf0 of pf0 technically first pf 
so vf0 is a pf

wizhaoredhat: Shouldn't you rely more on the PCI ID? Rather than the array indexes?

internal/daemon/vendor-specific-plugins/marvell/main.go
wizhaoredhat: @thom311 Is this why your VSP was failing on our OCP cluster?

thom311: I think so. Follow up at https://github.com/openshift/dpu-operator/pull/262

internal/daemon/vendor-specific-plugins/marvell/mrvl-utils/mrvlutils.go
wizhaoredhat: If your hardware team says this is ok, then we will just use the array index. But please do your best to see if we can use sysfs. This is potentially fragile to kernel code changes.

bn222: @thom311 has a valid remark. I recommend not making assumptions. Soon, we will have the ability to test without real hardware, and the order of devices returned by this function might not be the same.

internal/daemon/vendor-specific-plugins/marvell/debug-dp/debugdp.go
```diff
@@ -0,0 +1,41 @@
+package DebugDP
```
wizhaoredhat, line 1: Could you add a comment to this file what this is used for?

alkama-hasan: This is an example of adding other Data plane for marvell vsp in near future 
as of now this can be used to debug all the API call to data plane as we are planning to packages ovs-dpdk data plane for now as a seperate pod.
currently this functionality(debug dp) can be helpful for debugging mrvl-vsp without data plane it is only dummping the info called from vsp.

thom311: Yes, sounds useful.

But could you plase add what you just said as a comment? Your comment on github here cannot be found tomorrow, when somebody reviews the code.

It is also not clear to me how to use this in practice. Could you elaborate and give a short usage example?

alkama-hasan: @thom311 sure i will add in comment.
Regarding usage.
we have a flag in vsp(right now just a const inside main.go future plan isto make it configurable during run time) to choose dataplane where one can opt "debug" to use this funcitonality and this will dump all the everything to dataplane from vsp.
can be used for debug purpose of vsp without having a data plane also

internal/daemon/vendor-specific-plugins/marvell/mrvl-utils/mrvlutils.go
wizhaoredhat: In the interest of merging this PR. Could Marvell look at this on the side?

internal/daemon/vendor-specific-plugins/marvell/debug-dp/debugdp.go
wizhaoredhat: Please add these as a comment then I think we can merge this in.

internal/daemon/vendor-specific-plugins/marvell/ovs-dp/ovsdp.go
alkama-hasan: The code is generic can be used with or without dpdk

internal/daemon/vendor-specific-plugins/marvell/debug-dp/debugdp.go
alkama-hasan: done

PR #236, opened 2024-12-27
Rename ocpcluster masters

bn222:
testing if CI works again

### DIFF
--- a/hack/cluster-configs/config-dpu-host.yaml
+++ b/hack/cluster-configs/config-dpu-host.yaml
@@ -9,17 +9,17 @@ clusters:
     - name: "dpu_operator_host"
       dpu_operator_path: "../../"
     masters:
-    - name: "nicmodecluster-master-1"
+    - name: "{{cluster_name}}-master-1"
       kind: "vm"
       node: "localhost"
       ip: "192.168.122.2"
       disk_size: 32
-    - name: "nicmodecluster-master-2"
+    - name: "{{cluster_name}}-master-2"
       kind: "vm"
       node: "localhost"
       ip: "192.168.122.3"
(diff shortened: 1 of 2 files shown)


bn222:
@vrindle can you make this pass?

bn222:
/hold

Waiting for @vrindle to make sure our ci runs the necessary tests automatically

hack/cluster-configs/config-dpu-host.yaml
```diff
@@ -13,17 +13,17 @@ clusters:
       builder_image: "wsfd-advnetlab-amp04.anl.eng.bos2.dc.redhat.com:5000/ocp/builder:rhel-9-golang-1.22-openshift-4.18"
       base_image: "wsfd-advnetlab-amp04.anl.eng.bos2.dc.redhat.com:5000/ocp/4.18:base-rhel9"
     masters:
-    - name: "nicmodecluster-master-1"
+    - name: "{{cluster_name}}-master-1"
```
thom311, line 16: Why is this change desirable?

If you run the same config on different clusters, it seems nicer if the nodes have the same names. Otherwise, the name depends on the cluster, and if you are familiar with the name on Cluster A, you have to first find how they are called on Cluster B.

The commit message should explain why this is desirable.

bn222: nicmodecluster is a left-over that we forgot to clean up.

bn222: The intent of this PR is to test the CI because of the latest changes.

bn222:
/test make-e2e-test

Known flake that Intel is working on fixing

PR #241, opened 2025-01-07
Move to latest RHEL nightly on ACC

SalDaniele:


bn222:
This one is next, rebase needed though

PR #248, opened 2025-01-10
Run tests twice and custom CDA with no coldboot

bn222:


bn222:
/hold

we want to manually check if this worked as expected.

PR #253, opened 2025-01-14
Fix flake and bump golang container

bn222:


bn222:
can't bump builder images yet since the multi-arch version of the builder isn't ready yet.

PR #255, opened 2025-01-15
[th/cluster-configs-marvell-dpu] cluster-configs: add "config-marvell-dpu-{,host}.yaml"

thom311:
These configs are similar to the IPU ones, but are obviously for another cluster that has Marvell DPUs (Cluster 15 and 16).

### DIFF
--- a/hack/cluster-configs/config-marvell-dpu.yaml
+++ b/hack/cluster-configs/config-marvell-dpu.yaml
@@ -1,0 +1,23 @@
+# This deploys the Marvell DPU itself, installs microshift and the dpu-operator.
+# Run on hosts:
+# - wsfd-advnetlab44.anl.eng.bos2.dc.redhat.com (Cluster 15: Marvell dev cluster)
+# - wsfd-advnetlab41.anl.eng.bos2.dc.redhat.com (Cluster 16: Marvell dev cluster)
+clusters:
+  - name : "iso_cluster"
+    api_vip: "192.168.122.99"
+    ingress_vip: "192.168.122.101"
+    network_api_port: "eno2"
+    kind: "iso"
+    install_iso: "rhel:9.6"
+    masters:
+    - name: "marvell-dpu-{{worker_number(0)}}"
+      node: "{{worker_name(0)}}"
+      kind: "marvell-dpu"
+      ip: "192.168.123.7"
+    postconfig:
+    - name: "rh_subscription"
+      organization_id: "{{organization_id()}}"
+      activation_key: "{{activation_key()}}"
+    - name: "microshift"
+    - name: "dpu_operator_dpu"
+      dpu_operator_path: "../../"


hack/cluster-configs/config-marvell-dpu-host.yaml
```diff
@@ -0,0 +1,34 @@
...
+    workers:
+    - name: "worker-{{worker_number(0)}}"
+      kind: "physical"
+      node: "{{worker_name(0)}}"
+      bmc_user: "root"
+      bmc_password: "calvin"
+      bmc: "{{bmc(0)}}"
+    postconfig:
+    - name: "dpu_operator_host"
+      dpu_operator_path: "../../"
```
bn222, line 34: Can we reuse the same config?

hack/cluster-configs/config-marvell-dpu.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - name : "iso_cluster"
+    api_vip: "192.168.122.99"
+    ingress_vip: "192.168.122.101"
+    network_api_port: "eno2"
+    kind: "iso"
+    install_iso: "rhel:9.4"
+    masters:
+    - name: "marvell-dpu-{{worker_number(0)}}"
+      node: "{{worker_name(0)}}"
+      kind: "marvell-dpu"
```
bn222, line 15: This line is needed because the Marvell doesn't have a bmc?

thom311: No. It's need because CDA needs to know which kind of iso-cluster to deploy.

hack/cluster-configs/config-marvell-dpu-host.yaml
thom311: same as `config-dpu-host.yaml`?

Yes, after that IPU file gets a change `network_api_port: "{{api_network}}"`.

bn222: Can we try without explicitly specifying the network api port? It should be detected automatically.

thom311: I dropped the config file `config-mavell-dpu-host.yaml` from this pull request.

Instead, I will open a pull request that changes `config-dpu-host.yaml` to use `network_api_port: {{api_network()}}` so that it also works for Marvell clusters (no15 and no16).

hack/cluster-configs/config-marvell-dpu.yaml
thom311: > No. It's needed because CDA needs to know which kind of iso-cluster to deploy.

it's needed, because the deployoment for a iso-cluster works differently with Marvell DPU vs. IPU. Part of the reason for that is that Marvell doesn't have a BMC (yet) and we use pxeboot to install the Marvell DPU.

Hence, CDA needs to know which kind of approach to take, and it determines it based on the node kind `marvell-dpu` vs. `physical`.

hack/cluster-configs/config-marvell-dpu-host.yaml
thom311: opened PR #261 with this.

thom311:
I just repushed the patch to update the config to have `install_iso: "rhel:9.6"`

bn222:
error during install, stopped it manually since it hanged. Added check in CDA and bumped CDA.

retesting
/test make-e2e-test

thom311:
> error during install, stopped it manually since it hanged. Added check in CDA and bumped CDA.

This config only really works on the Marvell Clusters 15 and 16 (wsfd-advnetlab44 and wsfd-advnetlab41), where I successfully am running this exact configuration. The same as you can only meaningfully run `hack/cluster-configs/config-dpu.yaml` on the IPU clusters like Cluster 14-b (wsfd-advnetlab217).

The failure of the CI cannot have anything to do with this PR, since the CI does not use the new file (yet).

This PR should just be merged.

bn222:
correct, but it does need to pass CI regardless.

thom311:
closing. Apparently another approach will be chosen, where there is only one config `hack/cluster-configs/config-dpu.yaml` for all DPUs.

PR #262, opened 2025-01-20
[th/mrvl-enable-ipv6-link-local] rework configuring IPv6 link local addresses for Marvell VSP

thom311:
This is a follow up to https://github.com/openshift/dpu-operator/pull/233#discussion_r1920269782

Instead of using NetworkManager to configure link-local addresses, configure them only in kernel (via iproute2).


thom311:
CC: @alkama-hasan

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -17,6 +17,7 @@ import (
 	pb "github.com/openshift/dpu-operator/dpu-api/gen"
 	"github.com/openshift/dpu-operator/internal/utils"
 	opi "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
+	"github.com/pkg/errors"
```
bn222, line 19: Let's use `fmt.Errorf`

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -397,40 +398,25 @@ func getInterfaceName(deviceID string) (string, error) {
...
-	nmcliCmdStr := baseNsenterCmd + "nmcli con add type ethernet ifname " + interfaceName + " ipv6.method link-local"
-	nmcliCmd := exec.Command("/bin/sh", "-c", nmcliCmdStr)
-	if err := nmcliCmd.Run(); err != nil {
-		return err
-	}
+func enableIPV6LinkLocal(interfaceName string, ipv6Addr string) error {
+	// Ensure to set addrgenmode and toggle link state (which can result in creating
+	// the IPv6 link local address. Ignore errors here.
+	exec.Command("ip", "link", "set", interfaceName, "addrgenmode", "eui64").Run()
+	exec.Command("ip", "link", "set", interfaceName, "down").Run()
```
wizhaoredhat, line 405: Why do we set it down then up in the next line?

wizhaoredhat: Is this the only way to have the IPv6 address set?

wizhaoredhat, line 405: There is no better way then toggling the link to get an IPv6 link local address?

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -397,40 +398,25 @@ func getInterfaceName(deviceID string) (string, error) {
...
-	link, err := netlink.LinkByName(interfaceName)
+	err := exec.Command("ip", "link", "set", interfaceName, "up").Run()
 	if err != nil {
-		return err
+		return errors.Errorf("Error setting link %s up: %v", interfaceName, err)
 	}
 
-	//returns the list of IPv6 addresses on the given interface
-	address, err := netlink.AddrList(link, netlink.FAMILY_V6)
+	err = exec.Command("ip", "addr", "add", ipv6Addr+"/64", "dev", interfaceName).Run()
```
wizhaoredhat, line 412: Why do we need both an IPv6 link local and static IP address? Shouldn't we just have the static one?

bn222: I prefer the LL address

thom311: I don't think there is a problem.

A link local address is anyway only be usable and unique on the local network (that is, behind the interface on which they are configured).

You can configure addresses `fe80::2` on multiple interfaces just fine. Of course, whenever you use it, you must specify which interface you mean (e.g. `ping fe80::1%ens7f0`).

thom311:
one problem is that on the host NetworkManager is running, which (without configuration) will indefinitely try to activate a "Wired connection #" profile. It tries to activate the profile, failing to get a DHCP/autoconf address, and tearing the interface down. Thereby deleting the address.

I think the operator should issue a `nmcli device set $IFNAME managed no` before it starts to configure the interface.

thom311:
> one problem is that on the host NetworkManager is running, w

added a commit "marvell: run VSP as root and set interface unmanaged in NetworkManager"

internal/daemon/vendor-specific-plugins/marvell/main.go
bn222: @wizhaoredhat multiple DPUs is in the works and I plan to push a first PR for that. The change will be split up across multiple PRs. That's not only a Marvell specific use case. Is there a reason to target that use case this early (i.e. during this PR)?

wizhaoredhat: @thom311 I think that is fine then. Although specifying the interface would need to be added later on. I wonder if gRPC is ok to specify interfaces using that format e.g. fe80::1%ens7f0. Or will the socket needs to be bound to ens7f0

@bn222 I think this is ok to be considered now. Thomas' approach seems fine after his explanation.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -398,6 +398,9 @@ func getInterfaceName(deviceID string) (string, error) {
 // enableIPV6LinkLocal function to enable the IPv6 Link Local Address on the given Interface Name
 // It will return the error
 func enableIPV6LinkLocal(interfaceName string, ipv6Addr string) error {
+	// Tell NetworkManager to not manage our interface.
```
wizhaoredhat, line 401: Do we need nsenter? Shouldn't this be a host network pod instead?

internal/daemon/vendor-specific-plugins/marvell/main.go
thom311: I don't think you can meaningfully use IPv6 link local addresses without an interface name. Pretty sure you will always have to specify it, and any socket API is likely to support that.

internal/daemon/vendor-specific-plugins/marvell/main.go
thom311: hm. Maybe.

for one, currently the VSP container doesn't install /usr/bin/nmcli. But doing so might also not be desirable, because the NetworkManager version on the host might not be compatible with the nmcli version inside the container (in practice, it likely would work). We could instead mount parts of /usr/bin and /usr/lib (or all of /) inside the container, and try to use the host's nmcli inside the container. That sounds also cumbersome, to load the proper shared libraries (maybe a chroot would work).

Also, for nmcli to talk to D-Bus, we need to make `/var/run/dbus/system_bus_socket` accessible.

This probably could also be done.

But is there a problem with using nsenter?

bn222: Does it make sense to use netlink library instead? What's cost/benefit trade-off?

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -393,40 +397,28 @@ func getInterfaceName(deviceID string) (string, error) {
...
-func enableIPV6LinkLocal(interfaceName string) error {
-	baseNsenterCmd := "nsenter -t 1 -m -u -n -i -- "
-	nmcliCmdStr := baseNsenterCmd + "nmcli con add type ethernet ifname " + interfaceName + " ipv6.method link-local"
-	nmcliCmd := exec.Command("/bin/sh", "-c", nmcliCmdStr)
-	if err := nmcliCmd.Run(); err != nil {
-		return err
-	}
+func enableIPV6LinkLocal(interfaceName string, ipv6Addr string) error {
+	// Tell NetworkManager to not manage our interface.
+	exec.Command("nsenter", "-t", "1", "-m", "-u", "-n", "-i", "--", "nmcli", "device", "set", interfaceName, "managed", "no").Run()
```
alkama-hasan, line 402: what if this command does not excutes successfully?

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -393,40 +397,28 @@ func getInterfaceName(deviceID string) (string, error) {
...
-		return err
+		return fmt.Errorf("Error configuring IPv6 address %s/64 on link %s: %v", ipv6Addr, interfaceName, err)
 	}
 
-	if len(address) == 0 {
-		klog.Error("IPv6 address not found on interface", interfaceName)
-		return errors.New("IPv6 address not found on interface")
-	}
+	// Ping IPv6 Multicast group to get the IPv6 Neighbour Entry in Arp Table
+	exec.Command("ping6", "-c", "2", "-I", interfaceName, "ff02::1").Run()
```
alkama-hasan, line 420: will this command i.e. ping6 be able to add arp entry for the neighbour of this interface from inside pod?

internal/daemon/vendor-specific-plugins/marvell/main.go
thom311: But what to do about a failure? Erroring out seems overly strict.

It seems unnecessary trying to handle it. Maybe NetworkManager is not even running (though, I am not sure whether microshift/openshift has a strong dependency on NetworkManager).

internal/daemon/vendor-specific-plugins/marvell/main.go
thom311: The container runs already in the host's netns. I don't think there is a difference between `nsenter -t 1 -n ... ping6` and running `ping6` directly. Or what would be the difference?

PR #263, opened 2025-01-20
remove -a from build

bn222:
@SalDaniele quick fix for slow builds I came across while I'm adding incremental builds


wizhaoredhat:
@bn222 Your commit message should describe why the change was done.

bn222:
This is a very trivial change, but I've added an description either way. @wizhaoredhat ptal again.

PR #264, opened 2025-01-20
hack: bump cda

SalDaniele:
Add a workaround to avoid watchdog timeouts due to frequent ACC re-installation when developing / running CI jobs


bn222:
/lgtm

current flake should be addressed by Intel

PR #266, opened 2025-01-22
Bump intel

bn222:


bn222:
@SalDaniele ptal

Pulling in ip address cleanup

PR #268, opened 2025-01-22
Iic 427

SalDaniele:


SalDaniele:
@sudhar-krishnakumar PTAL

sudhar-krishnakumar:
@SalDaniele LGTM

SalDaniele:
Hi @sudhar-krishnakumar as discussed offline, jobs are failing because when the ACC comes up after RHEL install the devmen commands have not been run.

As mentioned, in subsequent reboots we do see the ACC comes up with connectivity.

We can merge this as is and leave the workaround to reboot the IMC after ACC install, but please make sure the solution address the case where the ACC installs RHEL and reboots

SalDaniele:
/retest

Failed due to go bug when building where go build hangs indefinitely

PR #269, opened 2025-01-23
Dockerfile.IntelVSP: Check-in pip dependencies

SalDaniele:
This container is built internally in a disconnected environment. We need these dependencies checked in locally.

Really we need these dependencies removed altogether, but as a temporary step we can make them available during the build process.


SalDaniele:
@sudhar-krishnakumar @saynb PTAL.

FYI this is why https://github.com/openshift/dpu-operator/pull/269 is needed. This image was flagged and unable to be built when we tried to onboard with Red Hat's official build systems because these dependencies could not be resolved in an offline environment.

It will be very challenging to maintain this moving forward, and poses security risks to have random binary files checked in like this. Lets aim to remove the python dependency altogether.

joepvd:
/hold
If this is intended for OCP product builds, binaries in the repo will not do.

bn222:
@sudhar-krishnakumar please note the comment from @joepvd 
I refer back to our discussion regarding python dependencies. I will bump up the priority of the jira that tracks getting rid of the python dependency

elfosardo:
@bn222 @SalDaniele you can have a look at how we build ironic containers if you need python dependencies in the downstream pipeline https://github.com/openshift/ironic-image
you don't have a lot so it should be manageable
we use a hybrid approach (RPMs + source code)

sudhar-krishnakumar:
> @sudhar-krishnakumar please note the comment from @joepvd I refer back to our discussion regarding python dependencies. I will bump up the priority of the jira that tracks getting rid of the python dependency

@bn222 , Sure Balasz. I think @saynb is looking at this.

PR #271, opened 2025-01-24
[th/hack-scripts-cleanup] cleanup hack scripts w.r.t. python and venv

thom311:
Some minor cleanup.

- hack: drop unnecessary "python3.11 -m venv /tmp/cda-venv"
- hack: call ./cda.py directly without explicit python interpreter
- hack: don't call versioned pip/python binaries

### DIFF
--- a/hack/both.sh
+++ b/hack/both.sh
@@ -3,7 +3,6 @@
 set -e
 
 cd cluster-deployment-automation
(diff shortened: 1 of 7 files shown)


bn222:
nice, thanks

/lgtm
/ok-to-test

PR #275, opened 2025-01-29
IIC 294 - Move to idpf v6.12

SalDaniele:
Bump RHEL to include idpf 6.12

With the move to idpf 6.12 we do not expect the host idpf to crash. It should be able to recover in the event the card reboots.

Move to latest RHEL 9.6 and remove WA in lockstep


bn222:
This is on the dpu side. On the host, we already bumped to 6.12 since we're deploying -nightly

PR #278, opened 2025-01-31
    IPv6 link local rework  to Marvell VSP

alkama-hasan:
This Commits replaces generated IPv6 address with static one for the link local 
    comm channel between host and dpu.

    Signed-off-by: Alkama Hasan <alkamah@marvell.com>

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"regexp"
 	"strconv"
-	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -205,7 +204,7 @@ func (vsp *mrvlVspServer) GetDeviceHealth(nfInterfaceName string) string {
 func (vsp *mrvlVspServer) Init(ctx context.Context, in *pb.InitRequest) (*pb.IpPort, error) {
 	klog.Infof("Received Init() request: DpuMode: %v", in.DpuMode)
 	vsp.isDPUMode = in.DpuMode
-	ipPort, err := vsp.fetchIP(in.DpuMode)
+	ipPort, err := vsp.configureIP(in.DpuMode)
 	if vsp.isDPUMode {
 		if vsp.deviceStore == nil {
 			vsp.deviceStore = make(map[string]mrvlDeviceInfo)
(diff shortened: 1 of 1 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"regexp"
 	"strconv"
-	"strings"
```
thom311, line 12: The commit message has leading indentation/whitespace. This is unusual.

thom311, line 12: in `enableIPV6LinkLocal()`, I think the `ping6 ... ff02::1` can be removed. Or is it still useful?

If yes, could you better explain in a code comment why it is still useful?

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -395,94 +394,41 @@ func (vsp *mrvlVspServer) SetNumVfs(ctx context.Context, in *pb.VfCount) (*pb.Vf
...
 		return pb.IpPort{}, err
 	}
 	klog.Infof("Interface Name: %s", ifName)
-
-	err = enableIPV6LinkLocal(ifName, IPv6AddrHost)
+	err = enableIPV6LinkLocal(ifName, addr)
+	addr = IPv6AddrDpu
 	if err != nil {
-		klog.Errorf("Error occurred in enabling IPv6 Link local Address: %v", err)
+		vsp.log.Error(err, "Error occurred in enabling IPv6 Link local Address: %v")
```
thom311, line 416: I think the change to use `vsp.log` is a good one. And you had that as a patch on some other pull request. It was then dropped, after I suggested to split it to a separate branch (I didn't mean to drop it entirely).

For now, I think we should consistently use `klog`. But still have another commit that moves all uses to `vsp.log`.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -395,94 +394,41 @@ func (vsp *mrvlVspServer) SetNumVfs(ctx context.Context, in *pb.VfCount) (*pb.Vf
...
+	var connStr string
+	if dpuMode {
+		connStr = "[" + addr + "%" + ifName + "]"
+	} else {
+		connStr = "[" + addr + "%25" + ifName + "]"
 	}
-	ConnStr := "[" + LinkLocalIpv6 + "%25" + ifName + "]"
-	klog.Infof("IPv6 Address: %s", LinkLocalIpv6)
-	klog.Infof("Connection String: %s", ConnStr)
+	klog.Infof("Connection String: %s", connStr)
```
thom311, line 426: I think this logging message and the previous one should be combined.

alkama-hasan: did you mean about connStr? connStr is different , and good to have in logging is what i believe
regarding the above line before and after enableipv6linklocal function call just logging in success case and before calling

internal/daemon/vendor-specific-plugins/marvell/main.go
alkama-hasan: yes not needed i will remove this thanks

internal/daemon/vendor-specific-plugins/marvell/main.go
alkama-hasan: yes i will have a seperate PR for logging change(from klog to vsp.log) may be at the end

internal/daemon/vendor-specific-plugins/marvell/main.go
thom311: I mean the code
```go
    klog.Infof("IPv6 Link Local Address Enabled  IfName: %v", ifName)
    var connStr string
    if dpuMode {
         connStr = "[" + addr + "%" + ifName + "]"
    } else {
         connStr = "[" + addr + "%25" + ifName + "]"
    }
    klog.Infof("Connection String: %s", connStr)
```
it logs two lines "IPv6 Link Local Address Enabled  IfName" and "Connection String". Those two lines should be combined, and the same information can be printed on one line.

thom311 (approved):

PR #284, opened 2025-02-03
Changes for p4 podification

saynb:
This includes changes for 
* `dnsPolicy: ClusterFirstWithHostNet` in vsp DS yaml
* go.mod update
* Calling VSP intel plugin with another option for P4 service name

Intel vendor changes
* Conversion of svc name to IP

Misc vendor changes
* Updating crypto, sys, term, text

Needs more ->
* oc create for the service
* oc create for the p4 pod


saynb:
@SalDaniele This PR is backward compatible when running P4 as a container. Can we run the CI?

SalDaniele:
LGTM for the most part.

One question, will this change to the vsp still work with the container based p4-container? Otherwise, we will need to modify CDA to stop deploying the container and deploy the p4 daemonset instead.

Eventually we will want to stop deploying the p4 container all together from CDA and deploy the p4 container from the vsp.

SalDaniele:
@saynb FYI e2e-test is failing due to a known flake in the CI. If you rebase on https://github.com/openshift/dpu-operator/pull/289 it will fix this issue, or wait till these PR goes in tomorrow.

saynb:
> One question, will this change to the vsp still work with the container based p4-container? Otherwise, we will need to modify CDA to stop deploying the container and deploy the p4 daemonset instead.

Yes I tested it on our cluster with the following variations

1. P4 as pod and vsp-p4-service up and running: It works as intended. `vsp-p4-service.default.svc.cluster.local` will be tried in VSP by default
2. P4 as container and service isn't up : `vsp-p4-service.default.svc.cluster.local` will be tried to be resolved. Since service isn't up, resolution will fail and we revert back to localhost:9559 which will be the same as old behavior
3. P4 as container and service is up: Haven't tested but should work since resolution will give service IP. Can test it out

SalDaniele:
/retest

Failed due to flake in go when cross compiling.

SalDaniele:
/retest

New flake that caused building on the host to fail

PR #285, opened 2025-02-03
Pull in sriov drain module and add test

SalDaniele:
We will re-use the drain interface provided by sriov network operator.
To start pulling in this code, and also validate behavior against a kind
cluster, add an appropriate test.

Pull in sriov drain module and add test


pkgs/drain/drain_test.go
```diff
@@ -0,0 +1,196 @@
...
+		// Create a KindCluster
+		testCluster = testutils.KindCluster{Name: "dpu-operator-test-cluster"}
+		restConfig = testCluster.EnsureExists()
+
+		k8sClient, err = client.New(restConfig, client.Options{})
+		Expect(err).NotTo(HaveOccurred())
+		node = getNode(k8sClient)
+	})
+	AfterAll(func() {
+		testCluster.EnsureDeleted()
```
bn222, line 134: This will break fast tests

pkgs/drain/drain_test.go
```diff
@@ -0,0 +1,196 @@
...
+			By("checking that all nodes are ready and schedulable")
+			ctrl.Log.Info("Checking node is ready", "nodeName", node.Name)
+
+			Expect(isNodeReady(node)).To(BeTrue(), "Node is not ready: "+node.Name)
+			Expect(isNodeSchedulable(node)).To(BeTrue(), "Node is not schedulable: "+node.Name)
+		})
+		It("should be able to run a workload", func() {
+			By("scheduling a test workload")
+			ctrl.Log.Info("Creating test pod on node", "nodeName", node.Name)
+			createTestPod(k8sClient)
```
bn222, line 147: func name

pkgs/drain/drain_test.go
```diff
@@ -0,0 +1,196 @@
...
+			Expect(drainSuccess).To(BeTrue())
+
+			ctrl.Log.Info("Checking node is cordoned", "nodeName", node.Name)
+			Expect(node.Spec.Unschedulable).To(BeTrue(), "Node is still schedulable after drain: "+node.Name)
+		})
+		It("should have any workload pods drained", func() {
+			By("verifying the test pod has been removed")
+			ctrl.Log.Info("Ensuring workload has been drained", "nodeName", node.Name)
+			Eventually(func() bool {
+				return testPodIsRunning()
```
bn222, line 173: `Eventually(testPodIsRunning, ...)` ?

SalDaniele:
/retest

Hit flake where init fails, seems like the underlying issue is the APFs are not created for some reason: https://issues.redhat.com/browse/IIC-514
 2025-02-14T02:22:11Z    ERROR   Daemon  Failed to listen        {"error": "Failed to get addr:port from VendorPlugin: Failed to start serving on grpcPlugin start: rpc error: code = Unknown desc = Error during init call, already initialized"}

SalDaniele:
/retest

Flake where registry.ci image is not available at build time

SalDaniele:
/retest

Acc seems to have randomly rebooted while vsp was coming up

pkgs/drain/drain_test.go
```diff
@@ -0,0 +1,215 @@
...
+	Context("When node drain and cordon has been requested", func() {
+		It("should drain the node and mark it unschedulable", func() {
+			By("checking that all nodes are not ready and draining")
+			Eventually(func() bool {
+				return testPodIsRunning()
+			}, testutils.TestAPITimeout, testutils.TestRetryInterval).Should(BeTrue(), "Pod did not become running in expected time")
+
+			ctrl.Log.Info("Draining node", "nodeName", node.Name)
+			var err error
+			testDrainer, err = createNewDrainer(restConfig)
```
bn222, line 179: I think we need a drain.go file that forwards most of the requests to sriov drain implementation. Somewhat like a facade pattern. That will allow us to reuse createNewDrainer elsewhere in this repo too.

pkgs/drain/drain_test.go
```diff
@@ -0,0 +1,215 @@
...
+var (
+	restConfig    *rest.Config
+	k8sClient     client.Client
+	testCluster   testutils.KindCluster
+	node          corev1.Node
+	testDrainer   drain.DrainInterface
+	podName       = "test-nginx-pod"
+	containerName = "test-nginx-container"
+	podNamespace  = "default"
+	imageName     = "nginx:latest"
```
bn222, line 31: Bit confusing to call the image nginx

pkgs/drain/drain_test.go
```diff
@@ -0,0 +1,215 @@
...
+			Eventually(func() bool {
+				return testPodIsRunning()
+			}, testutils.TestAPITimeout, testutils.TestRetryInterval).Should(BeTrue(), "Pod did not become running in expected time")
+
+			ctrl.Log.Info("Draining node", "nodeName", node.Name)
+			var err error
+			testDrainer, err = createNewDrainer(restConfig)
+			Expect(err).NotTo(HaveOccurred())
+
+			drainSuccess, err := testDrainer.DrainNode(context.TODO(), &node, true)
```
bn222, line 182: lines 172 to 181, could we move it above 171 so that the context in our BDD framework here makes sense?

pkgs/drain/drain_test.go
SalDaniele: I added this, PTAL and let me know if this is what you had in mind

pkgs/drain/drain_test.go
SalDaniele: Done

SalDaniele:
@bn222 All feedback has been addressed, ready for review

bn222:
To clarify, it's simply the string `nginx` that could cause some confusion in this context.

SalDaniele:
@bn222 PTAL, needs new lgtm, had to update some rbac rules

PR #291, opened 2025-02-06
Changes to pick Intel PR for removal of python deps(https://github.com/intel/ipu-opi-plugins/pull/334)

sudhar-krishnakumar:
This PR adds the p4runtime python deps, and shell
script-> p4runtime-2023.11.0/copy_p4rt_python_deps.sh to copy them over in VSP's Dockerfile, as per intel PR->
https://github.com/intel/ipu-opi-plugins/pull/334
This PR adds the dependencies under here in DPU repo: https://github.com/openshift/dpu-operator/tree/main/cmd/intelvsp.
As a follow-up, would need change to update VSP's Dockerfile managed in DPU
repo(https://github.com/openshift/dpu-operator/blob/main/Dockerfile.IntelVSP.rhel).
For reference, there is Intel's VSP Dockerfile here(that no longer uses p4runtime wheel packag):
https://github.com/intel/ipu-opi-plugins/blob/main/ipu-plugin/images/Dockerfile


bn222:
@sudhar-krishnakumar Can you make the changes to the Dockerfile too?

sudhar-krishnakumar:
> @sudhar-krishnakumar Can you make the changes to the Dockerfile too?

@bn222 We dont have access to proprietary RHEL base image, to test the change.

bn222:
@sudhar-krishnakumar could you nevertheless push a pr with what you think should be changed in the dockerfile in this repo? If it passes out ci, it is in a good enough state. If you want to take small steps and merge the current pr first, then by all means, but we would need your help on making the changes in the dockerfile too

SalDaniele:
/close

Added Dockerfile in https://github.com/openshift/dpu-operator/pull/292

sudhar-krishnakumar:
> @sudhar-krishnakumar could you nevertheless push a pr with what you think should be changed in the dockerfile in this repo? If it passes out ci, it is in a good enough state. If you want to take small steps and merge the current pr first, then by all means, but we would need your help on making the changes in the dockerfile too

@bn222 we can test with centos image as base. But without using your RHEL Image, we could break the VSP and CI pipeline can fail. Are you okay with that. I am just bringing this up, based on your prior feedback. Preferrably, we dont want to push something, that we cannot test with your base image. We wont be able to test without access to proprietary RHEL image. Are you okay, if it breaks.

sudhar-krishnakumar:
@SalDaniele Looks like you have closed this, and created new one with updated VSP Dockerfile.

PR #296, opened 2025-02-07
Refactor

bn222:


internal/daemon/daemon.go
```diff
@@ -47,7 +47,7 @@ func NewDaemon(mode string, client client.Client, vspImages map[string]string, c
 	}
 }
 
-func (d *Daemon) Run() error {
+func (d *Daemon) ListenAndServe() error {
```
SalDaniele, line 49: Is this a little confusing? Now the daemon will have 2 methods ListenAndServer differentiated only by the return values IIUC?

Maybe daemon.ListenAndServer() should be  renamed StartListenAndServer or something more specific.

Or Run->InitDaemon or something.

internal/daemon/hostsidemanager.go
```diff
@@ -268,9 +267,7 @@ func (d *HostSideManager) Serve(listener net.Listener) error {
 
 func (d *HostSideManager) Stop() {
```
SalDaniele, line 278: General question, who do who expect to invoke a sideManager's stop() method? Makes sense for the daemon to clean up after itself, but trying to understand when in the sequence this would be called.

internal/daemon/daemon.go
bn222: The refactor is not complete yet. The goal is to have two functions:

Listen
and
Serve.

Listen always returns a socket that needs to be passed to Serve. Any pre-init required before Serve blocks is done in Listen so that the design becomes composable. That means that you will be able to use mulitple SideManagers in Daemon. Daemon itself will also have a Listen and a Serve. The intent remains the same, call Listen recursively inside the top level Listen of daemon to init all SideManagers returning one of the sockets and pass to back in to Serve.

Following this design, it will be possible to test the hardware detection (and the DPU CRs) in a race-condition free way.

internal/daemon/hostsidemanager.go
bn222: Stop() is callable if you don't use ListenAndServe. It only makes sense to invokte stop if Serve is in a go-routine. If we have multiple side managers, if one errors out, we want to error out the whole binary. For that, we will call stop on all the other SideManagers that have not errored out to stop all the go routines in which we run Serve.

PR #297, opened 2025-02-10
GetDevices() API rework to Marvell VSP

alkama-hasan:
This commits add support for GetDevices() on host side to Marvell VSP

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"regexp"
 	"strconv"
-	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -205,7 +204,7 @@ func (vsp *mrvlVspServer) GetDeviceHealth(nfInterfaceName string) string {
 func (vsp *mrvlVspServer) Init(ctx context.Context, in *pb.InitRequest) (*pb.IpPort, error) {
 	klog.Infof("Received Init() request: DpuMode: %v", in.DpuMode)
 	vsp.isDPUMode = in.DpuMode
-	ipPort, err := vsp.fetchIP(in.DpuMode)
+	ipPort, err := vsp.configureIP(in.DpuMode)
 	if vsp.isDPUMode {
 		if vsp.deviceStore == nil {
 			vsp.deviceStore = make(map[string]mrvlDeviceInfo)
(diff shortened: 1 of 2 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -53,10 +54,11 @@ type mrvldp interface {
 	DeleteDataplane(bridgeName string) error
 }
 type mrvlDeviceInfo struct {
-	nfInterfaceName string
-	dpInterfaceName string
-	dpMAC           string
-	health          string
+	secInterfaceName string
```
wizhaoredhat, line 57: Why the rename to `secInterfaceName` ?

alkama-hasan: well this is no more only nf interface now .the same deviceStore is being used for host side also in that case it will be an sdp vf (which will be  also used for sec network)
so secInterface suits good here is what i believe

thom311: Could you please split the patch in a trivial first commit (that does the renaming) and a non-trivial second commit that does the API rework?

Or is the renaming the "API rework"? Instead of calling it "API rework", could you better describe in the commit message what the patch does? Renaming an internal variable does not really count as an API rework, and I don't see what API otherwise gets reworked.

internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -174,16 +177,26 @@ func (vsp *mrvlVspServer) ConfigureNetworkInterface() error {
...
+		nfLink, err := netlink.LinkByName(secInterfaceName)
+		if err != nil {
+			return "Unhealthy"
+		}
+		//check if the interface is up =0 means interface is down
+		if nfLink.Attrs().Flags&net.FlagUp == 0 {
+			return "Unhealthy"
+		}
+		return "Healthy"
+	case "sriov":
```
thom311, line 198: aren't both cases exactly the same? Use `case "veth", "sriov": ...`

internal/daemon/vendor-specific-plugins/marvell/mrvl-utils/mrvlutils.go
```diff
@@ -19,6 +19,29 @@ const (
...
+			targetAddresses = append(targetAddresses, dev.Address)
+		}
+	}
+	if len(targetAddresses) == 0 {
+		return nil, errors.New("no devices found with given Vendor ID and Device ID")
+	}
+	var interfaceNames []string
+	for _, addr := range targetAddresses {
+		interfaceName, _ := GetNameByPCI(addr)
+		interfaceNames = append(interfaceNames, interfaceName)
```
thom311, line 40: I think you should check for error in `GetNameByPCI()`, and if so, skip the `interfaceName`.

If you do that, drop the `if len(targetAddresses) == 0` check above, and add a `if len (interfaceNames) == 0` check below.

thom311: and maybe just merge the two for loops into one. There is no need to first build a list of `targetAddresses` before resolving them to Name. This could be done in one loop and drop the list `targetAddresses`.

thom311 (commented):
I think the commit message should explain in more detail what the commit does.

internal/daemon/vendor-specific-plugins/marvell/mrvl-utils/mrvlutils.go
bn222: I think having a separate loop is cleaner since each loop does one thing. Also, move each loop into a function and add the names `buildTargetAddresses` and `resolveToPciAddresses`.

wizhaoredhat:
@bn222 Can we remove your hold on this PR?

internal/daemon/vendor-specific-plugins/marvell/mrvl-utils/mrvlutils.go
alkama-hasan: as discussed i will remove the second loop as now have return list of pciAddress instead of name

internal/daemon/vendor-specific-plugins/marvell/main.go
alkama-hasan: Yes i think renaming of this variable is part of API rework(Not exactly GetDevices API ) as the scope of this API on VSP now changed to serve both host as well as DPU i have changed this name from nf_interface to sec_interface to serve this purpose .
yes i think i can write a better commit message

internal/daemon/vendor-specific-plugins/marvell/main.go
alkama-hasan: Yes both are same i think it;s good to merge this together thanks,

PR #298, opened 2025-02-10
[th/cluster-config-network-port] cluster-configs: set "network_api_port" for ISO clusters from cluster info

thom311:
We want to use "hack/cluster-configs/config-dpu.yaml" to deploy also
our Marvell DPUs.

In our clusters, we will need "network_api_port" set to "eno2" instead
of "eno12409".

Get the value from the cluster info in the Google spread sheet.


bn222:
@vrindle is this a failure due to us?

vrindle:
it seemed the test timed out. running again...

PR #299, opened 2025-02-11
Refactor

bn222:
This increases testing coverage and makes it possible to instantiate multiple side managers. A few bugs that were revealed by increased testing are fixed as well.


bn222:
@SalDaniele I need a bit more work on this to fix the flake this has uncovered.

bn222:
/test make-e2e-test
forgot to add new cda-config.yaml on this server

PR #301, opened 2025-02-12
[th/config-install-iso] config: update install_iso for "config-dpu.yaml"

thom311:
We need a more recent rhel-9.6 kernel for Marvell DPU (and host side). Update the iso.


wizhaoredhat:
LGTM

SalDaniele:
@thom311 FYI this is failing because we checkout the commit for make test rather than merge it on main. I will push a fix to this and retry

PR #303, opened 2025-02-14
marvell: package CP agent with VSP

bbhamidipati:
This commit adds support to compile and run Octeon CP agent with Marvell VSP.


internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,122 @@
...
+setup_hugepages() {
+    if ! mount | grep -q "^none on /dev/huge type hugetlbfs " ; then
+        mkdir -p /dev/huge
+        mount -t hugetlbfs none /dev/huge
+        echo 1024 > /proc/sys/vm/nr_hugepages
+    fi
+}
+
+setup_dpu_link() {
+    # from http://file.brq.redhat.com/~thaller/SDP-CP-agent.docx
```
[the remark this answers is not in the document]
bbhamidipati: Removed URL.

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/sh
+
+set -x
+
+load_driver() {
+	chk=$(lsmod | grep -w octeon_ep)
+	if [[ -z $chk ]]
```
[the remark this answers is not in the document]
bbhamidipati: Replaced.

Dockerfile.mrvlVSP.rhel
```diff
@@ -26,5 +29,41 @@ RUN yum update -y \
...
+
+ENV CFLAGS="-Wno-error=unused-result -DUSE_PEM_AND_DPI_PF=1"
+RUN if [ "$TARGETARCH" = "arm64" ] ; then \
+        git clone https://github.com/MarvellEmbeddedProcessors/pcie_ep_octeon_target.git && \
+        cd /workspace/pcie_ep_octeon_target/target/libs/octep_cp_lib && make ; \
+    fi
+
+ENV OCTEP_PATH="/workspace/pcie_ep_octeon_target/target/libs/octep_cp_lib"
+ENV CFLAGS="$CFLAGS -I/workspace/bin/include -I$OCTEP_PATH/include"
+ENV LDFLAGS="-L/workspace/bin/lib -L$OCTEP_PATH"
```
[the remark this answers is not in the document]
bbhamidipati: Fixed.

Dockerfile.mrvlVSP.rhel
```diff
@@ -16,6 +18,7 @@ RUN mkdir -p /bin && \
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
 ARG TARGETARCH
+
 COPY --from=builder /workspace/bin/vsp-mrvl.${TARGETARCH} vsp-mrvl
 ENV DEBIAN_FRONTEND=noninteractive
```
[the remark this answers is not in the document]
bbhamidipati: Done.

Dockerfile.mrvlVSP.rhel
```diff
@@ -26,5 +29,41 @@ RUN yum update -y \
        make \
        iputils \
        iproute
-USER 65532:65532
-ENTRYPOINT ["/vsp-mrvl"]
+
+RUN if [ "$TARGETARCH" = "arm64" ] ; then \
+        dnf update -y && dnf install patch kmod tar gawk make gcc g++ git pciutils -y ; \
```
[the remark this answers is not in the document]
bbhamidipati: I added dnf install go while using a different FROM image, and its not needed.
I think no need to change code around builder.

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,120 @@
...
+}
+
+unload_driver() {
+    rmmod octeon_ep_vf -f
+    chk=$(lsmod | grep -w octeon_ep)
+    if [ -n $chk ]
+    then
+    rmmod octeon_ep -f
+    echo "Unloaded octeon_ep and octeon_ep_vf drivers, wait for 20 seconds before retry"
+    sleep 20
```
thom311, line 30: intentation.

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,120 @@
...
+    path="/sys/bus/pci/devices/$pf/net/"
+    ifname=$(ls $path)
+
+    state=$(cat /sys/class/net/$ifname/operstate)
+    if [ $state == "up" ]
+    then
+        echo "PF found: $ifname"
+        return 0
+    else
+        return -1
```
thom311, line 45: The valid return codes range from 0 to 255. -1 will be mapped to 255. This is unnecessary confusing.

Just `return 1`. Or `return 255`, if you really want that.

bbhamidipati: Done

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
bbhamidipati: done

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,122 @@
...
+        then
+        setup_host_link
+        elif test $arch = "aarch64"
+        then
+	setup_dpu_link
+	if [[ $? -ne 0 ]]; then
+		echo "Failed to set up CP Agent"
+		return
+	fi
+        fi
```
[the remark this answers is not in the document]
bbhamidipati: Done

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
```diff
@@ -26,6 +26,13 @@ spec:
           runAsUser: 0
         command: {{.Command}}
         args: {{.Args}}
+        resources:
```
wizhaoredhat, line 29: Could you explain why this is needed? The hugepages?

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
```diff
@@ -35,6 +42,20 @@ spec:
...
+        - mountPath: /lib/modules/
+          mountPropagation: HostToContainer
+          name: host-libmodules
+        - mountPath: /dev/
+          mountPropagation: Bidirectional
+          name: host-dev
+        - mountPath: /sys/bus/pci/
+          mountPropagation: Bidirectional
+          name: host-sysbuspci
+        - mountPath: /tmp/
```
wizhaoredhat, line 54: Why do you need to mount tmp? Can't you use the tmp folder inside the container?

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
bbhamidipati: CP Agent uses hugepages.

Dockerfile.mrvlVSP.rhel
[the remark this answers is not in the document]
bbhamidipati: Fixed.

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
bbhamidipati: Host' tmp is not required. Fixed same.

bn222:
This has an effect on how the general vendor vsp works. If the test passes, I'm fine to add it for the time being.

wizhaoredhat:
> This has an effect on how the general vendor vsp works. If the test passes, I'm fine to add it for the time being.

Good that you caught this. I was thinking this too

@bbhamidipati Would it be possible to add more jinja substitutions to make it possible for the vendor to customize those options?

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
```diff
@@ -26,6 +26,13 @@ spec:
           runAsUser: 0
         command: {{.Command}}
         args: {{.Args}}
+        resources:
+          requests:
+            memory: 100Mi
+            hugepages-2Mi: 100Mi
+          limits:
+            memory: 100Mi
+            hugepages-2Mi: 100Mi
```
thom311, line 35: see https://issues.redhat.com/browse/RHEL-78129?focusedId=26827364&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-26827364

Quote Kiet Tran:

> Marvell DPU (cn10k) device requires a size of hugepages of multiple of 4MB for aligned base addresses.
>
> Since RHEL kernel has only support for sizes at 2MB then next size up of 32MB, we advised RH to configure kernel when booting and allocate hugepages at size of 32MB when launching octeon_cp_agent on target DPU for a working solution. No code change is needed.

I think you need to request `hugepages-32Mi`.

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
```diff
@@ -38,6 +45,12 @@ spec:
         - mountPath: /proc
           mountPropagation: Bidirectional
           name: host-proc
+        - mountPath: /lib/modules/
```
thom311, line 48: I think the container should not try to mix paths from the host with the ones from the container.

I suggest to not mount this path, and instead run `chroot /host modprobe octeon_ep`. Or where else is `/lib/modules` needed?

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
bbhamidipati: Addressed

Dockerfile.mrvlVSP.rhel
```diff
@@ -29,5 +62,11 @@ RUN yum update -y \
...
-USER 65532:65532
-ENTRYPOINT ["/vsp-mrvl"]
+
+WORKDIR /workspace
+
+USER 0
+
+COPY internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh /
+RUN ["chmod", "+x", "/setup_vsp.sh"]
+ENTRYPOINT ["/bin/sh", "-c", "/setup_vsp.sh"]
```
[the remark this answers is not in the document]
bbhamidipati: Done

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/bash
```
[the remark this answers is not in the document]
bbhamidipati: Done

internal/daemon/vendor-specific-plugins/marvell/setup_vsp.sh
```diff
@@ -0,0 +1,120 @@
...
+    exec /usr/bin/octep_cp_agent /usr/bin/cn106xx.cfg -- --dpi_dev $dpi --pem_dev $pem &> /tmp/octep-cp-log.txt &
+
+    return 0
+
+}
+
+
+run() {
+    arch=$(uname -m)
+    if test $arch = "x86_64"
```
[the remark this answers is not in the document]
bbhamidipati: Done

Dockerfile.mrvlVSP.rhel
```diff
@@ -29,5 +62,11 @@ RUN yum update -y \
...
        iproute
-USER 65532:65532
-ENTRYPOINT ["/vsp-mrvl"]
+
+WORKDIR /workspace
+
+USER 0
+
+COPY internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh /
+RUN ["chmod", "+x", "/vsp-mrvl-run.sh"]
```
thom311, line 71: This is unnecessary. The script is already marked as executable.

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
```diff
@@ -0,0 +1,125 @@
...
+        mode="dpu"
+	echo "mode: dpu"
+    else
+	mode="host"
+	echo "mode: host"
+    fi
+
+    if test $mode = "host" ; then
+        setup_host_link
+    elif test $mode = "dpu"; then
```
thom311, line 114: - don't mix style regarding indentation (tabs vs. spaces).
- don't mix style regarding `if [ ... ]` vs. `if test ...`.
- in `test $mode = "dpu"`, the `$mode` should be quoted, the sane string literal `"dpu"` can optionally not be quoted.

```bash
    if [ "$(lspci -d 177d:a0f7)" != "" ] ; then
        mode="dpu"
    else
        mode="host"
    fi

    echo "mode: $mode"

    if [ "$mode" = "host" ] ; then
        setup_host_link
    elif [ "$mode" = "dpu" ] ; then
        setup_dpu_link
```

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
```diff
@@ -0,0 +1,125 @@
...
+    then
+        chroot /host modprobe octeon_ep
+        sleep 1
+        echo "Loaded octeon_ep driver"
+    fi
+
+    chk=$(lsmod | grep -w octeon_ep_vf)
+    if [ -z $chk ]
+    then
+        chroot /host modprobe octeon_ep_vf
```
thom311, line 17: I don't think it's necessary or useful to load `octeon_ep_vf` module. That module gets automatically loaded when `sriov_numvfs` gets set (and if `octeon_ep` module is connected otherwise).

Likewise, there is no need to `rmmod octeon_ep_vf` below. It gets automatically unloaded when `octeon_ep` gets unloaded.


Currently the script only supports a single Marvell DPU in the host. That is fine for now. In the future, the script should load `octeon_ep` once (and keep it loaded). And then iterate over all device, and re-bind the driver where `detect_link` indicates that is necessary.

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
```diff
@@ -0,0 +1,125 @@
...
+        sleep 20
+    fi
+}
+
+detect_link() {
+    pf="0000:$(lspci -d 177d:b900 -n | awk 'NR==1{print $1}')"
+    path="/sys/bus/pci/devices/$pf/net/"
+    ifname=$(ls $path)
+
+    state=$(cat /sys/class/net/$ifname/operstate)
```
thom311, line 39: What do you want to check here?

operstate is a combination of lowerup (carrier) and administratively up. Kernel does not detect carrier unless you set the interface administratively up (`ip link set ... up`). And, since you just loaded the module, nobody is expected to set this interface up for you (maybe NetworkManager will do it, but you shouldn't rely on that).

I think, you need first issue a `ip link set "$ifname" up`. And after (some timeout?) check whether there is carrier on the interface.

(also, throughout the script, I think variables should be consistently quoted in almost all cases)

bbhamidipati: sequence:
dpu: cp agent is up and running

host: 
load octeon_ep module
check dmesg for fw ready status 1 ==> pf interface gets created and comes up naturally

I am just relying on this to make sure link is detected.

Dockerfile.mrvlVSP.rhel
```diff
@@ -15,12 +19,41 @@ COPY . .
...
 # Use distroless as minimal base image to package the Marvell VSP binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
 ARG TARGETARCH
+
 COPY --from=builder /workspace/bin/vsp-mrvl.${TARGETARCH} vsp-mrvl
-ENV DEBIAN_FRONTEND=noninteractive
+COPY --from=builder /workspace/bin/lib/libconfig* /usr/lib64/
+COPY --from=builder /workspace/pcie_ep_octeon_target/target/apps/octep_cp_agent/cn106xx.cfg /usr/bin/
+COPY --from=builder /workspace/pcie_ep_octeon_target/target/apps/octep_cp_agent/bin/bin/octep_cp_agent /usr/bin/
```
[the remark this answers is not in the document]
bbhamidipati: This is not needed, already being copied here:
RUN if [ "$TARGETARCH" = "arm64" ] ; then \
        export OCTEP_PATH="/workspace/pcie_ep_octeon_target/target/libs/octep_cp_lib" && \
        export CFLAGS="$CFLAGS -I/workspace/bin/include -I$OCTEP_PATH/include" && \
        export LDFLAGS="-L/workspace/bin/lib -L$OCTEP_PATH" && \
        make && \
        cp bin/bin/octep_cp_agent /usr/bin/ && \
        cp cn106xx.cfg /usr/bin/ && \
        cp /workspace/bin/lib/libconfig* /usr/lib64/; \
    fi

Dockerfile.mrvlVSP.rhel
bbhamidipati: removed

internal/daemon/plugin/bindata/vsp-ds/99.vendor-plugin-daemonset.yaml
bbhamidipati: done, fixed

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
bbhamidipati: done

Dockerfile.mrvlVSP.rhel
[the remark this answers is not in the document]
bbhamidipati: curl -OL is added

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
thom311: I see. When the PF interface gets created, it is already `UP`. That seems unusual behavior from the driver, but OK then.

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
thom311: ok. Is it relevant to unload that too?

Dockerfile.mrvlVSP.rhel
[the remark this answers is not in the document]
thom311: > This is not needed, already being copied here:

what you quote happens in the `builder` container/stage.

I don't think that will cause the files to end up in the final container...

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
bbhamidipati: Yes, we need to unload vf driver.

bn222:
cp agent should be shipped inside the container, not as a systemd service

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
```diff
@@ -0,0 +1,84 @@
...
+            unload_driver
+            echo "Unloaded octeon_ep and octeon_ep_vf drivers, wait for 20 seconds before retry"
+            sleep 20
+        fi
+    done
+}
+
+setup_dpu_link() {
+    cp /cp-agent.service /host/etc/systemd/system/cp-agent.service
+    chroot /host systemctl enable cp-agent
```
thom311, line 56: you need a `chroot /host systemctl daemon-reload` after changing unit files on disk.

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
```diff
@@ -0,0 +1,11 @@
...
+Description=cp_agent container
+
+[Service]
+Restart=always
+ExecStart=/usr/bin/podman run --rm --name cp_agent --privileged --network host -v /tmp:/tmp -v /lib/modules:/lib/modules  -v /sys/bus/pci:/sys/bus/pci -v /:/host -v /dev:/dev --entrypoint /cp-agent-run.sh cp-agent-rhel
+
+ExecStop=/usr/bin/podman stop cp_agent
+
+[Install]
+WantedBy=local.target
```
thom311, line 11: there is no `local.target` on RHEL installation (is there)?

I think the needs `WantedBy=multi-user.target`

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=cp_agent container
```
thom311, line 2: Could you improve the service description? That the service runs a container, is just an implementation detail. Maybe "Octeon Control Plane Agent for Marvell Octeon Device"?

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
```diff
@@ -0,0 +1,11 @@
...
+
+[Service]
+Restart=always
+ExecStart=/usr/bin/podman run --rm --name cp_agent --privileged --network host -v /tmp:/tmp -v /lib/modules:/lib/modules  -v /sys/bus/pci:/sys/bus/pci -v /:/host -v /dev:/dev --entrypoint /cp-agent-run.sh cp-agent-rhel
```
thom311, line 6: - `cp-agent-rhel` is not a generally available URL where to find the container. During deployment of dpu-operator, we build the container. This URL needs to be configured and passed down all the way.
- I think the `--entrypoint` should not be overwritten.
- `--pull newer`, otherwise, the service fails if there currently is no internet.

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
```diff
@@ -0,0 +1,84 @@
...
+NUM_VFS=8
+
+load_driver() {
+    chroot /host modprobe octeon_ep
+    chroot /host modprobe octeon_ep_vf
+}
+
+unload_driver() {
+    chroot /host rmmod octeon_ep_vf -f
+    chroot /host rmmod octeon_ep -f
```
thom311, line 14: why `-f`?. That seems wrong.

For one, you wouldn't want to force a removal. Also, RHEL is not compiled with `CONFIG_MODULE_FORCE_UNLOAD`, so it has no effect anyway.

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
thom311: to make it configurable, do
```
Environment=IMAGE={{ image }}
ExecStart=/usr/bin/podman run ... ${IMAGE}
```

Here, the `{{ image }}` must come all the way from the upper layers.

That way, I can `systemctl edit cp-agent.service` and set a different IMAGE value (at least for testing).

Dockerfile.mrvlCPAgent.rhel
```diff
@@ -0,0 +1,79 @@
...
+# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
+
+# Due to https://github.com/golang/go/issues/70329 cross-compilation hangs at times.
+# As a temporary workaround, we can try specifying GOMAXPROCS=2 to relieve this issue
+RUN mkdir -p /bin && \
+    GOMAXPROCS=2 CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} make build-marvell-vsp
+
+WORKDIR /workspace
+RUN if [ "$TARGETARCH" = "arm64" ] ; then \
+        curl -OL -k https://hyperrealm.github.io/libconfig/dist/libconfig-1.7.2.tar.gz && \
```
bn222, line 26: no internet during build so no curl

internal/daemon/vendor-specific-plugins/marvell/cp-agent-run.sh
```diff
@@ -0,0 +1,66 @@
...
+	echo "mode: dpu"
+        setup_dpu_link
+        if [ $? -ne 0 ]; then
+            echo "Failed to set up CP Agent"
+            return
+        fi
+    fi
+}
+
+run
```
bn222, line 66: could move this into go

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
```diff
@@ -0,0 +1,84 @@
...
+        if [ $? -ne 0 ]; then
+            echo "Failed to set up CP Agent"
+            return
+        fi
+    fi
+
+    /vsp-mrvl
+}
+
+run
```
bn222, line 84: more stuff that can go into go. We don't want to repeat the problems other projects have where this gets out of hand and we end up with an operator written in sh.

bn222 (commented):
@thom311 Here are the comments from a first pass.

Dockerfile.mrvlCPAgent.rhel
thom311: @bn222 other containers already calls `pip install`, so there needs to be internet.

Anyway. If this were changed, what is your suggestion to avoid it? For example:
- compile the binary files and commit to git (and `COPY` over). That would work, since those binaries have few dependencies, so if you build against a reasonably old glibc, then it will also work in newer RHEL containers.
- build the files in a container, push them to e.g. `quay.io/user/dpu-operator-build-dep` and use it via

```
FROM quay.io/user/dpu-operator-build-dep:octep_cp_agent as builder_cp_agent
COPY --from builder_cp_agent ...
```

I think this should be done in a followup commit.

bn222: The other containers that use `pip install` use cachito. That's not set up for this external repo, nor would that work for curling in.

I propose that we create an rpm, add it to our repos, and then use `dnf install` which uses a local vetted version of the RHEL repos.

bn222: It needs to be done in this commit, or this image will not be built by ART and it will not be available.

thom311: How about (the last 2 commits of) https://github.com/thom311/dpu-operator/commits/th/tmp1/ ?

bn222: short term this is ok for me. At some point, we need to either move away from this dependency, or add it to the RHEL repos.

thom311: @bbhamidipati could you rebase on top of https://github.com/thom311/dpu-operator/commits/th/tmp1/ (if you agree)?

Also, RHEL actually has `libconfig-devel` package. I think we should use that instead of recompiling the library from source. Or does that not work? Let me try to do that at https://github.com/wizhaoredhat/marvell-octeon-10-tools/ ...

bbhamidipati:
Work in progress

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
bbhamidipati: done

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
bbhamidipati: done

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
bbhamidipati: done.
Makefile does:
copies internal/daemon/vendor-specific-plugins/marvell/cp-agent/cp-agent.service   /bin/cp-agent.service
appends Image from Makefile

Tested, works fine w/o editing.

Dockerfile.mrvlCPAgent.rhel
bbhamidipati: works with libconfig-devel

Dockerfile.mrvlCPAgent.rhel
```diff
@@ -0,0 +1,78 @@
...
+
+# Due to https://github.com/golang/go/issues/70329 cross-compilation hangs at times.
+# As a temporary workaround, we can try specifying GOMAXPROCS=2 to relieve this issue
+
+WORKDIR /workspace
+RUN if [ "$TARGETARCH" = "arm64" ] ; then \
+        tar -xvf internal/daemon/vendor-specific-plugins/marvell/cp-agent/libconfig-1.7.2.tar.gz -C . && \
+        cd libconfig-1.7.2 && ./configure --prefix=/workspace/bin/ && \
+        make && make install && cd - && \
+        rm -fr /workspace/libconfig-1.7.2* ; \
```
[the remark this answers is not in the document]
bbhamidipati: libconfig-devel is working fine

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
bbhamidipati: moved into go

internal/daemon/vendor-specific-plugins/marvell/vsp-mrvl-run.sh
bbhamidipati: done

internal/daemon/vendor-specific-plugins/marvell/cp-agent.service
bbhamidipati: done

taskfile.yaml
```diff
@@ -83,6 +83,18 @@ tasks:
...
+    vars:
+      IMAGE: '{{.IMAGE}}'
+      CPIMAGE: "Environment=IMAGE='{{.IMAGE}}'"
+      CPFILE: 'internal/daemon/vendor-specific-plugins/marvell/cp-agent/cp-agent.service'
+    cmds:
+    - mkdir -p bin
+    - echo "[Service]" > bin/cp-agent.service
+    - echo '{{.CPIMAGE}}' >> bin/cp-agent.service
+    - cat '{{.CPFILE}}' >> bin/cp-agent.service
+
```
bn222, line 97: not here

taskfile.yaml
```diff
@@ -128,6 +148,9 @@ tasks:
       - task: ginkgo
       - task: envtest
     cmds:
+      - task: build-cpagent-service
+        vars:
+          IMAGE: '{{.REGISTRY}}:5000/mrvl-cpagent:dev'
```
bn222, line 153: not here

taskfile.yaml
bbhamidipati: moved

taskfile.yaml
bbhamidipati: Passing Image as argument with build-image-all. 

If thats not fine, we can add like below, and access directly in images.yaml?

+++ taskfile.yaml
vars:
  DPU_OPERATOR_TEMP_DIR: /tmp/dpu-operator-tmp
  BINDIR: bin
  REGISTRY:
    sh: hostname
  ENVTEST_K8S_VERSION: 1.27.1
+  CPAGENTIMAGE: '{{.REGISTRY}}:5000/mrvl-cpagent:dev'

bn222: Why are we adding code that when you run it creates a script here instead of adding the script somewhere into the repo?

bbhamidipati: script needs image url as configurable

bn222: place the script from the container, and move the logic from the build system into go. Then add

`TODO: move away from using systemd file`

bbhamidipati: done

bbhamidipati:
work in progress

Dockerfile.mrvlCPAgent.rhel
```diff
@@ -0,0 +1,59 @@
...
+
+WORKDIR /workspace
+RUN GOMAXPROCS=2 CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -o /usr/bin/cp-agent-run internal/daemon/vendor-specific-plugins/marvell/cp-agent/cp-agent-run.go
+
+# Use distroless as minimal base image to package the Marvell CP agent binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN dnf config-manager --set-enabled crb
```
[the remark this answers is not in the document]
bbhamidipati: removed

Dockerfile.mrvlCPAgent.rhel
```diff
@@ -0,0 +1,59 @@
...
+# Due to https://github.com/golang/go/issues/70329 cross-compilation hangs at times.
+# As a temporary workaround, we can try specifying GOMAXPROCS=2 to relieve this issue
+
+WORKDIR /workspace
+RUN GOMAXPROCS=2 CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -o /usr/bin/cp-agent-run internal/daemon/vendor-specific-plugins/marvell/cp-agent/cp-agent-run.go
+
+# Use distroless as minimal base image to package the Marvell CP agent binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
+ENV DEBIAN_FRONTEND=noninteractive
```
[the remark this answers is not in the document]
bbhamidipati: removed

Dockerfile.mrvlCPAgent.rhel
```diff
@@ -0,0 +1,56 @@
...
+# the GOARCH has not a default value to allow the binary be built according to the host where the command
+# was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
+# the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
+# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
+
+# Due to https://github.com/golang/go/issues/70329 cross-compilation hangs at times.
+# As a temporary workaround, we can try specifying GOMAXPROCS=2 to relieve this issue
+
+WORKDIR /workspace
+RUN GOMAXPROCS=2 CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -o /usr/bin/cp-agent-run internal/daemon/vendor-specific-plugins/marvell/cp-agent/cp-agent-run.go
```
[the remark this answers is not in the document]
bbhamidipati: > COPY --from builder /usr/bin/cp-agent-run /usr/bin/cp-agent-run
This fails as base image already have this binary, and both are same.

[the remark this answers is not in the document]
thom311: where are the commits 
- `marvell: vendor "pcie_ep_octeon_target" sources`
- `marvell: add "vendor.sh" script for including "pcie_ep_octeon_target"`

were they dropped by accident?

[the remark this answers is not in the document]
thom311: > Its a two stage build.

Hm, ok, if you say so.

But I don't think my image has a `/usr/bin/cp-agent-run` file (on aarch64). Investigating.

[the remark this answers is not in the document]
thom311: > Its a two stage build.
>
> builder builds a dev-arm64 image. This will have /usr/bin/cp-agent-run.
>
> second stage will use this dev-arm64 as base image. This will also have /usr/bin/cp-agent-run.

How? The second stage starts with `FROM registry.ci.openshift.org/ocp/4.19:base-rhel9`.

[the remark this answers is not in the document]
thom311: regardless what incremental.go does, it must be possible to use the plain dockerfile with `buildah build`.

This is of course significantly complicated because (I think) you don't have access to registry.ci.openshift.org...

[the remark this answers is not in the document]
bn222: We have ci/prow/images to check if that works. Until it's fully enabled, we would need to add a target to our CI that uses `build-image-all`. @thom311 can you do that? That will automatically gate incoming incomplete PRs (not water tight though).

Dockerfile.mrvlCPAgent.rhel
```diff
@@ -0,0 +1,56 @@
...
+FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
+
+RUN yum update -y \
+    && yum install -y \
+       net-tools \
+       kmod \
+       pciutils \
+       iputils \
+       iproute \
+       libconfig-devel \
```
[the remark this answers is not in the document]
bbhamidipati: Fixed

Dockerfile.mrvlCPAgent.rhel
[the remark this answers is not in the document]
bbhamidipati: fixed the Dockerfile and tested, cp-agent-run works fine

Dockerfile.mrvlCPAgent.rhel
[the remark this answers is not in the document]
thom311: please mark completed discussion threads as done (I don't have permission to do that).

internal/daemon/vendor-specific-plugins/marvell/vendor/vendor.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
thom311, line 1: The script lacks the executable bit. Please `chmod +x`.

PR #304, opened 2025-02-14
[th/mrvl-optimistic-ipaddr] marvell: configure IPv6 addresses with optimistic DAD

thom311:
When adding IPv6 addresses, they are for a short time in tentative state while doing DAD. At that time, the address is not yet usable.

This causes failures:
```
Daemon  Failed to listen        {"error": "Failed to start listening on [fe80::1%enP2p1s0v0]:8085: listen tcp [fe80::1%enP2p1s0v0]:8085: bind: cannot assign requested address"}
```
Make the address optimistically available.

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"net"
+	"os"
 	"os/exec"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -441,6 +442,8 @@ func enableIPV6LinkLocal(interfaceName string, ipv6Addr string) error {
 		klog.Infof("nmcli device set %s managed no failed with error %v", interfaceName, err1)
 	}
 
+	_ = os.WriteFile("/proc/sys/net/ipv6/conf/"+interfaceName+"/optimistic_dad", []byte("1"), os.ModeAppend)
```
wizhaoredhat, line 445: Can we use the return code and log an error?

PR #306, opened 2025-02-15
Handle both primiary and secondary network traffic on FXP phy port0

arunvelayutham:
Redhat wanted to use single physical port to handle both their primary and secondary network traffic.
1) handle ARP/Broadcast packets using FXP mirroring to primary interface
   VSI(D5) while the original packets are sent to P0's VSI(D4).
2) Unicast packets are selectively forwarded to the primary network
   interface based on DMAC lookup.

NOTE: Default behavior will be defined using devmem to reroute all the phy0 traffic to D5's VSI. Once the VSP comes in, it will program the above mentioned mirroring and dmac lookup P4 rules to take precedence over the devmem defined default behavior.


SalDaniele:
Thanks for the PR @arunvelayutham.

I will need to manually update the system connections and re-build the p4 image to test this PR. Will update you if it succeeds.

PR #307, opened 2025-02-18
Set Owner Reference in VSP

SalDaniele:
Set vsp owner reference to dpuOperatorConfig

We want the vsp daemonset to be owned by some resource such that the vsp will be garbage collected properly.

Ideally this would be the daemon or dpu-daemon daemonset, but kubernetes does not allow pods or daemonsets to own other controllers.

A good step in the right direction is to have the vsp owned by the same custom resource that owns the dpu daemon.

A potenital next step might be to add an additional manager for coordinating deployment of the vsp.

### DIFF
--- a/internal/controller/dpuoperatorconfig_controller.go
+++ b/internal/controller/dpuoperatorconfig_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/dpu-operator/api/v1"
 	"github.com/openshift/dpu-operator/pkgs/render"
+	"github.com/openshift/dpu-operator/pkgs/vars"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -113,7 +114,7 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 func (r *DpuOperatorConfigReconciler) createCommonData(cfg *configv1.DpuOperatorConfig) map[string]string {
 	// All the CRs will be in the same namespace as the operator config
 	data := map[string]string{
-		"Namespace":              "openshift-dpu-operator",
+		"Namespace":              vars.Namespace,
 		"ImagePullPolicy":        r.imagePullPolicy,
 		"Mode":                   "auto",
 		"DpuOperatorDaemonImage": r.dpuDaemonImage,
--- a/internal/daemon/dpusidemanager_test.go
+++ b/internal/daemon/dpusidemanager_test.go
@@ -69,9 +69,10 @@ var _ = g.Describe("DPU side maanger", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 		}()
 
-		dpuPlugin := plugin.NewGrpcPlugin(true,
+		dpuPlugin, err := plugin.NewGrpcPlugin(true,
 			client,
 			plugin.WithPathManager(pathManager))
+		Expect(err).NotTo(HaveOccurred())
 		dpuDaemon = NewDpuSideManger(dpuPlugin, config, WithPathManager(pathManager))
 
 		dpuListen, err := dpuDaemon.Listen()
--- a/internal/daemon/plugin/vendorplugin.go
+++ b/internal/daemon/plugin/vendorplugin.go
@@ -78,7 +78,7 @@ type GrpcPlugin struct {
 func NewVspTemplateVars() VspTemplateVars {
 	return VspTemplateVars{
 		VendorSpecificPluginImage: "",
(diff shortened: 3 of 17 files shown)


SalDaniele:
@bn222 PTAL

internal/daemon/plugin/vendorplugin.go
```diff
@@ -136,7 +137,7 @@ func (gp *GrpcPlugin) deployVsp() {
 
 	if vspImage != "" {
 		gp.log.Info("Deploying VSP", "vspImage", vspImage, "command", gp.vsp.Command, "args", gp.vsp.Args)
-		err := render.ApplyAllFromBinData(gp.log, "vsp-ds", gp.vsp.ToMap(), binData, gp.k8sClient, nil, nil)
+		err := render.ApplyAllFromBinData(gp.log, "vsp-ds", gp.vsp.ToMap(), binData, gp.k8sClient, nil, scheme.Scheme)
```
bn222: How did this work before?

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -117,6 +117,7 @@ func (r *DpuOperatorConfigReconciler) createCommonData(cfg *configv1.DpuOperator
 		"ImagePullPolicy":        r.imagePullPolicy,
 		"Mode":                   "auto",
 		"DpuOperatorDaemonImage": r.dpuDaemonImage,
+		"DpuOperatorConfigName":  cfg.Name,
```
bn222, line 120: There is only one config name allowed (see https://github.com/openshift/dpu-operator/pull/213). In that case, we don't need to pass it in through a yaml file. In short, instead of allowed arbitrary names for this CR, we will ask everyone to use the same name since there is no reason to allow anything else in my view.

internal/daemon/dpusidemanager_test.go
```diff
@@ -69,9 +69,10 @@ var _ = g.Describe("DPU side maanger", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 		}()
 
-		dpuPlugin := plugin.NewGrpcPlugin(true,
+		dpuPlugin, err := plugin.NewGrpcPlugin(true,
 			client,
 			plugin.WithPathManager(pathManager))
+		Expect(err).NotTo(HaveOccurred())
```
bn222, line 75: nice

internal/daemon/plugin/vendorplugin.go
```diff
@@ -132,21 +133,37 @@ func WithVsp(template_vars VspTemplateVars) func(*GrpcPlugin) {
...
-	} else {
+	// It is not mandatory that a vsp image is provided. If not, we can assume this will be handled by the user and still return a GrpcClient
+	if vspImage == "" {
 		gp.log.Info("WARNING: VSP Image not set, skipping vendor plugin container startup")
+		return nil
+	}
+
+	// Retrieve the Dpu Operator Config which owns the Dpu Daemonset so we can ensure the vsp shares the same owner reference.
+	namespace := os.Getenv("NAMESPACE")
+	dpuOperatorConfigName := os.Getenv("DPU_OPERATOR_CONFIG_NAME")
```
bn222, line 147: Can we get rid of this?

internal/daemon/dpusidemanager_test.go
```diff
@@ -69,9 +69,10 @@ var _ = g.Describe("DPU side maanger", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
```
bn222, line 69: iiuc, you're making the DpuOperatorConfig be the owner for the vsp directly, not chaining through the Daemon because that won't work?

internal/daemon/plugin/vendorplugin.go
SalDaniele: Now that I know we can assume the config name, yes. I suppose we could assume the namespace as well, but typically this is something I see provided in the downward API, so I figured it would be safest to consume it.

internal/daemon/dpusidemanager_test.go
SalDaniele: Yeah exactly. I tried chaining through the daemon first, but after some trial / error and reading up on the docs, it seems kubernetes does not allow pods or daemonsets to own other controllers. It has to be a higher level resource (i.e. a custom resource).

internal/daemon/plugin/vendorplugin.go
bn222: I think we might need to create a struct called DownwardApi and add functions to fetch anything that's coming from the downward api. Why? Because code that calls os.Getenv is untestable. It's tedious to refactor code that is sprinkled with os.Getenv, and we should aim to avoid writing any code that ties ourselves to the os.Getenv. 

That being said, I think we don't need the downward api here at all, because we can just assume the hard coded name.

SalDaniele:
@bn222 All feedback has been addressed, ready for review

PR #308, opened 2025-02-19
Add gomaxproc when building

SalDaniele:
There is an issue that causes go multiarch to hang indefinitely sometimes [1].

Per the comments on the issue, it seems setting GOMAXPROCS to a small number can relieve this issue as a temporary workaround

[1] https://github.com/golang/go/issues/70329


thom311:
this lgtm. I agree with this approach.

PR #310, opened 2025-02-24
tft: disable hwol plugin

SalDaniele:
We are not enabling ovs-hwol, there is no reason to be running this plugin.

### DIFF
--- a/hack/cluster-configs/ocp-tft-config.yaml
+++ b/hack/cluster-configs/ocp-tft-config.yaml
@@ -17,8 +17,6 @@ tft:
         client:
           - name: "$worker"
             sriov: "true"
-        plugins:
-          - name: validate_offload
         secondary_network_nad: "default-sriov-net"
         resource_name: "openshift.io/dpu"
 kubeconfig: "/root/kubeconfig.ocpcluster"


SalDaniele:
IIUC, we will not see a vf-rep for non-switchdev vfs, so I think we need to remove the plugin.

This fails every time for the IPU too, but it looks like the tool does return a non-zero status for plugin failures, which is the only reason this hasnt been caught earlier.

thom311:
> This fails every time for the IPU too, but it looks like the tool does return a non-zero status for plugin failures, which is the only reason this hasnt been caught earlier.

The idea is that you run `print_results.py` on the output. Test one will fail on any failed tests (aside from pretty-printing the results).

The success exit state of tft.py indicates that the test ran to completion, not that all tests passed. I guess, we could add an option `--exit-code` to validate and fail the tool at the end...

SalDaniele:
@bn222 PTAL

SalDaniele:
@thom311 Understood, updated the PR to also properly evaluate the flow test results.

hack/traffic_flow_tests.sh
```diff
@@ -28,3 +28,5 @@ temp_file=$(mktemp)
 envsubst < ../hack/cluster-configs/ocp-tft-config.yaml > $temp_file
 
 ./tft.py $temp_file
+
+./print_results.py ft-logs/*RESULTS
```
thom311, line 32: the `*RESULTS` file is deprecated. See [here](https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/commit/1d03e96142a1b24ac1bbd728ebad140251ace641). Use `ft-logs/*.json`.

Maybe even:
```bash
OUTPUT_BASE="./ft-logs/result-$(date '+%Y%m%d-%H%M%S.%s')"

./tft.py --output-base "$OUTPUT_BASE" "$temp_file"

./print_results.py "$OUTPUT_BASE"*.json
```

SalDaniele:
/test make-e2e-test

Prow never ran the job for some reason..

SalDaniele:
@wizhaoredhat PTAL. Small change, just disabling an un-used test and ensuring we evaluate tft results properly

SalDaniele:
@vrindle PTAL this is failing without ever running.

The job is getting stuck in queue: 
image

SalDaniele:
/retest

Flake with google api being down

hack/traffic_flow_tests.sh
thom311: @SalDaniele  `./print_results.py ft-logs/*.json` works on the Jenkins workspace, which gets wiped every time.

however, it does not work when you run `make traffic-flow-tests` repeatedly on your machine (without wiping the directory before run). Then result files pile up in `kubernetes-traffic-flow-tests/ft-logs/`. Instead, we only want to compare the last run.

That was the point of `$OUTPUT_BASE` above.

thom311: let me open a PR

PR #314, opened 2025-02-28
Pull Intel Pr 349 for cli_client concurrency issue

sudhar-krishnakumar:
Pull Intel Pr 349 for cli_client concurrency issue.
Added cli_client related binaries, that fixes the concurrent
access issue, not available in current 1.8 release.
Also made change in VSP's Dockerfile. I have tested the changes  with podman container setup. Our cluster is under use for P4 podification work. Will need the CI/CDA run to validate(Dockerfile for Intel VSP change) and ipu-plugin changes in redhat setup.


Dockerfile.IntelVSP.rhel
```diff
@@ -28,6 +28,9 @@ RUN mkdir -p /opt/p4/p4-cp-nws/bin/p4
 COPY ./cmd/intelvsp/fxp-net_linux-networking.pkg /
 COPY ./cmd/intelvsp/p4rt-ctl /opt/p4/p4-cp-nws/bin/
 
+RUN mkdir -p /ipu-mgmtd
+COPY ./cmd/intelvsp/ipu-mgmtd /ipu-mgmtd
```
SalDaniele, line 32: Can you elaborate on what these binaries are? Is it not an option to push source code and compile as part of the Dockerfile build process?

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
```diff
@@ -908,6 +862,30 @@ fi
 
 }
 
+func preInitAppScript() string {
```
SalDaniele, line 865: We are overwriting the pre-init app? Is it possible to avoid this?

We rely on the pre-init-app to apply workarounds required to ensure Redfish functions properly: https://github.com/bn222/cluster-deployment-automation/blob/main/ipu.py#L175.

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
```diff
@@ -908,6 +862,30 @@ fi
...
+
+if [ -d "$WORKDIR" ]; then
+    cd $WORKDIR
+    if [ -e load_custom_pkg.sh ]; then
+        # Fix up the cp_init.cfg file
+        ./load_custom_pkg.sh
+    fi
+fi
+cd $CURDIR
+python /usr/bin/scripts/cfg_acc_apf_x2.py
```
SalDaniele, line 880: Is this needed now?

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
```diff
@@ -896,9 +850,9 @@ if [ -e %s ]; then
     sed -i 's/pf_mac_address = "00:00:00:00:03:14";/pf_mac_address = "%s";/g' $CP_INIT_CFG
     sed -i 's/acc_apf = 4;/acc_apf = %s;/g' $CP_INIT_CFG
     sed -i 's/comm_vports = .*/comm_vports = (([5,0],[4,0]),([0,3],[5,3]),([0,2],[4,3]));/g' $CP_INIT_CFG
-    sed -i 's/uplink_vports = .*/uplink_vports = ([0,0,0],[0,1,1],[4,1,0],[4,5,1],[5,1,0],[5,2,1]);/g' $CP_INIT_CFG
-    sed -i 's/rep_vports = .*/rep_vports = ([0,0,0],[4,5,1]);/g' $CP_INIT_CFG
-    sed -i 's/exception_vports = .*/exception_vports = ([0,0,0],[4,5,1]); /g' $CP_INIT_CFG
+    sed -i 's/uplink_vports = .*/uplink_vports = ([4,5,0],[5,1,0],[5,2,1]);/g' $CP_INIT_CFG
```
SalDaniele, line 853: Can you help me understand what we are achieving by changing these? What does this change do?

SalDaniele (commented):
It seems like we are:
pre-compiling binaries, baking these into the vsp and installing these on the IMC.

It feels like we are taking a round-about path to update the IMC firmware through the ACC because the IMC lacks certain upgrades / state that we need.

Wouldn't it make more sense to get these changes added to the MeV image itself, and pre-flash the card with these changes, rather than adding them at runtime.

Or is the idea that these changes are temporarily added for the sake of fixing the issue in the short-term, and in the long term will be part of the MeV release?

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
SalDaniele: I don't think this is a blocker if we can't work around this, but it makes running Redfish in the future a bit trickier.

sudhar-krishnakumar:
> It seems like we are: pre-compiling binaries, baking these into the vsp and installing these on the IMC.
> 
> It feels like we are taking a round-about path to update the IMC firmware through the ACC because the IMC lacks certain upgrades / state that we need.
> 
> Wouldn't it make more sense to get these changes added to the MeV image itself, and pre-flash the card with these changes, rather than adding them at runtime.
> 
> Or is the idea that these changes are temporarily added for the sake of fixing the issue in the short-term, and in the long term will be part of the MeV release?

On IMC we have a daemon process that runs, ps -aux, will show you ipumgmtd. cli_client tool is a client that gets the info from this daemon->ipumgmtd. There was a bug in the daemon, during concurrent access, and that got resolved. But this fix is not part of 1.8 release, expected to be part of 2.0 release. Until then(short-term) rolled in the fixed version of ipumgmtd into VSP. So VSP can install it during pre-init-step. pre-init-step is the ONLY place...where this can get copied over. Hope this clarifies.

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
sudhar-krishnakumar: @SalDaniele This is part of Arun's(@arunvelayutham ) single port change. It maps ACC(4)'s D5(5) interface to uplink port->P0 on IPU.

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
sudhar-krishnakumar: @SalDaniele Will clean it up later.

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
sudhar-krishnakumar: @SalDaniele I didnt know you were making changes here for redfish. With this change in VSP in pre-init for cli-client issue, it will over-write the pre-init , with the version in VSP.

Dockerfile.IntelVSP.rhel
sudhar-krishnakumar: I have explained it in my other command as far as what these binaries are. Build process is quite complex for this source code.

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
SalDaniele: Yeah was more asking for my own curiosity.

So the syntax is:

[ <number indicating component>, <number indicating interface num>, <num indicating p0/p1>] >

Is that correct? 4 = ACC, but I could map IMC / host interfaces here too?

SalDaniele:
> > It seems like we are: pre-compiling binaries, baking these into the vsp and installing these on the IMC.
> > It feels like we are taking a round-about path to update the IMC firmware through the ACC because the IMC lacks certain upgrades / state that we need.
> > Wouldn't it make more sense to get these changes added to the MeV image itself, and pre-flash the card with these changes, rather than adding them at runtime.
> > Or is the idea that these changes are temporarily added for the sake of fixing the issue in the short-term, and in the long term will be part of the MeV release?
> 
> @SalDaniele On IMC we have a daemon process that runs, ps -aux, will show you ipumgmtd. cli_client tool is a client that gets the info from this daemon->ipumgmtd. There was a bug in the daemon, during concurrent access, and that got resolved. But this fix is not part of 1.8 release, expected to be part of 2.0 release. Until then(short-term) rolled in the fixed version of ipumgmtd into VSP. So VSP can install it during pre-init-step. pre-init-step is the ONLY place...where this can get copied over. Hope this clarifies.


Ok this does clarify. We are currently doing something similar for the hotfix of redfish, where we pre-install the binary on the IMC, and in the pre-init app replace the bad redfish with the 2.0 version. See https://github.com/bn222/cluster-deployment-automation/blob/main/ipu.py#L177

Can we just copy these binaries over to our servers on our own, and install in the pre-init app, rather than checking them into production / customer facing repo and installing them here?

In my mind, hacks / workarounds like this should be performed by the user/admin (CDA) but ideally dpu operator should not do these sort of hacks if it is avoidable.

vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/ipuplugin/lifecycleservice.go
sudhar-krishnakumar: Yes. You are correct. 0-> x86 Host. 4-> ACC. 5-> IMC.

SalDaniele:
/close

A refactor of this was merged in https://github.com/openshift/dpu-operator/pull/316

PR #319, opened 2025-03-04
Adding changes for perms for daemonsets, roles and services

saynb:


SalDaniele:
Nice LGTM, thanks for the PR @saynb

PR #320, opened 2025-03-04
[th/mrvl-create-veth-pair-once] marvell: accept existing interface in createVethPair()

thom311:
This should should be idempotent and accept that the interface might already exist. Especially, since createVethPair() can be called multiple times.
```
  I0214 10:43:39.226702       1 main.go:218] Received Init() request: DpuMode: true
  ...
  I0214 10:43:39.542460       1 main.go:161] Creating Veth Pairs: 2
  ...
  I0214 10:44:06.202503       1 main.go:218] Received Init() request: DpuMode: true
  ...
  I0214 10:44:06.517237       1 main.go:161] Creating Veth Pairs: 2
  E0214 10:44:06.517348       1 main.go:165] Error occurred in creating Veth Pair: file exists
  I0214 10:44:06.719772       1 main.go:147] Deleted Veth Pair: nf_interface0
  I0214 10:44:06.919443       1 main.go:147] Deleted Veth Pair: nf_interface1
  E0214 10:44:06.919460       1 main.go:227] Error occurred in configuring Network Interface: file exists
```
It was indicated that the VSP should recognize when it already called createVethPair() earlier and skip that. That may be a solution, however, it does not conflict with createVethPair() sensibly handling the case where the interfaces were already created and making the function idempotent.

---

CC: @alkama-hasan

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -88,23 +88,35 @@ func (vsp *mrvlVspServer) createVethPair(index int) error {
 	//dpInterfaceName is the name of the interface on the Data Plane side
 	secInterfaceName := fmt.Sprintf("nf_interface%d", index)
 	dpInterfaceName := fmt.Sprintf("dp_interface%d", index)
-	vethLink := &netlink.Veth{
-		LinkAttrs: netlink.LinkAttrs{Name: secInterfaceName},
-		PeerName:  dpInterfaceName,
-	}
-	if err := netlink.LinkAdd(vethLink); err != nil {
-		return err
+
+	var nfLink netlink.Link
+	var peerLink netlink.Link
+	var err error
+
+	nfLink, _ = netlink.LinkByName(secInterfaceName)
+	if nfLink != nil {
+		peerLink, _ = netlink.LinkByName(dpInterfaceName)
 	}
 
-	nfLink, err := netlink.LinkByName(secInterfaceName)
-	if err != nil {
-		return err
+	if nfLink == nil || peerLink == nil {
+		vethLink := &netlink.Veth{
+			LinkAttrs: netlink.LinkAttrs{Name: secInterfaceName},
+			PeerName:  dpInterfaceName,
+		}
+		if err := netlink.LinkAdd(vethLink); err != nil {
+			return err
+		}
+		nfLink, err = netlink.LinkByName(secInterfaceName)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -88,25 +88,37 @@ func (vsp *mrvlVspServer) createVethPair(index int) error {
...
-		PeerName:  dpInterfaceName,
-	}
-	if err := netlink.LinkAdd(vethLink); err != nil {
-		return err
+
+	var nfLink netlink.Link
+	var peerLink netlink.Link
+	var err error
+
+	nfLink, _ = netlink.LinkByName(secInterfaceName)
```
SalDaniele, line 96: Why don't we check err on LinkByName?

Isn't it possible this fails on the by LinkByName()call to "execGetLink()".  Might be easier for debugging to catch/report this error earlier?

thom311: I think we strongly rely that the API guarantees `(nfLink != nil) == (err == nil)`?

If we would not rely on such conditions, we would always need to double check
```go
    if err != nil {
         return err
    }
    if nfLink == nil {
        return errors.New("no device returned")
    }
```
(which in general we don't do, because it's redundant and because we really want an error reason when the function fails to provide a result).

Note that below there is the check `if nfLink == nil || peerLink == nil {`, so we do check for error still. We just don't care about the actual `err` variable.

Maybe that's unclear and I should rework this to check for `err` instead. Is that what you suggest?

SalDaniele: I think it is fine, it just seems like we would lose any unexpected error message returned by netlink.LinkByName()

SalDaniele: Maybe I am misunderstanding as well...so then we expect that this function will fail some of the time, and if it fails for either the nfLink or Peer link, we know we should try to create the vethLink again? Is that correct?

thom311: the function first tries to fetch `nfLink` and `peerLink`.

If that succeeds, they are both taken.

If that fails (`if nfLink == nil || peerLink == nil {`), we create the veth pair. This is a check for error, it's just we don't care about the actual `error` variable, because there is nothing to be done about that.

> it just seems like we would lose any unexpected error message returned by netlink.LinkByName().

Those errors are not used. They are not lost, they are intentionally ignored.  Note that we do check for error  (`if nfLink == nil || peerLink == nil {`), just not for the `error` variable.

PR #327, opened 2025-03-13
Extend E2E test suite to validate Network Functions

SalDaniele:
In order to validate Network Function connectivity, we first need to validate pod-to-pod connectivity.

We will then validate SFC behavior, and validate that pod-to-pod still works once the Network Function has been created

We need to validate that Network Function l3 connectivity is working. Add the tests to validate connectivity both from Network Function to workload pod, and from external client to workload pod.

TODO: Tests assume the external client is the host running e2e-test-suite. It would be nice if an arbitrary client reachable from the DPU could be provided instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,16 @@ test: podman-check manifests generate fmt vet envtest ginkgo
 .PHONY: fast-test
 fast-test: envtest ginkgo
 	FAST_TEST=true KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) $(if $(TEST_FOCUS),-focus $(TEST_FOCUS),) ./internal/... ./pkgs/... ./api/v1/... -coverprofile cover.out
-##@ Build
+
+# Set default values for environment variables for the external client in e2e-test-suite
+NF_INGRESS_IP ?= 10.20.30.2
+EXTERNAL_CLIENT_DEV ?= eno12409
+EXTERNAL_CLIENT_IP ?= 10.20.30.100
+
+export NF_INGRESS_IP
+export EXTERNAL_CLIENT_DEV
+export EXTERNAL_CLIENT_IP
+
 .PHONY: e2e-test-suite
 e2e-test-suite: envtest ginkgo
 	FAST_TEST=true KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) $(if $(TEST_FOCUS),-focus $(TEST_FOCUS),) ./e2e_test/... -coverprofile cover.out
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -9,6 +9,8 @@ import (
 	g "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -19,8 +21,8 @@ import (
 	"github.com/openshift/dpu-operator/internal/scheme"
 	"github.com/openshift/dpu-operator/internal/testutils"
 	"github.com/openshift/dpu-operator/pkgs/vars"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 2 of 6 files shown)


SalDaniele:
@thom311 @wizhaoredhat PTAL

Makefile
```diff
@@ -207,7 +207,16 @@ test: podman-check manifests generate fmt vet envtest ginkgo
...
-##@ Build
+
+# Set default values for environment variables for the external client in e2e-test-suite
+NF_INGRESS_IP ?= 10.20.30.2
+EXTERNAL_CLIENT_DEV ?= eno12409
+EXTERNAL_CLIENT_IP ?= 10.20.30.100
+
+export NF_INGRESS_IP
+export EXTERNAL_CLIENT_DEV
+export EXTERNAL_CLIENT_IP
```
thom311, line 218: I think the parent commit is not going to work without this change, is it? The commit should be squashed.

---

But the actual problem is that this won't work on some Marvell Clusters. For example, on wsfd-advnetlab44 this interface name is going to be `eno2`. Before merging, it should be made sure that this also works (at least conceptionally) with Marvell, otherwise, this just creates work for later. A possible solution can also be to just skip skip an entire test on a certain backend/DPU, except `dpu-operator` is currently (intentionally) not aware which DPU kind it is running on...

e2e_test/e2e_test.go
```diff
@@ -387,10 +387,10 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 			fmt.Println("Creating workload pods")
 			Eventually(func() bool {
 				return testutils.PodIsRunning(hostSideClient, testPodName, "default")
-			}, testutils.TestAPITimeout*30, testutils.TestRetryInterval).Should(BeTrue(), "Pod did not become running in expected time")
+			}, testutils.TestAPITimeout*45, testutils.TestRetryInterval).Should(BeTrue(), "Pod did not become running in expected time")
 			Eventually(func() bool {
 				return testutils.PodIsRunning(hostSideClient, testPod2Name, "default")
-			}, testutils.TestAPITimeout*30, testutils.TestRetryInterval).Should(BeTrue(), "Pod did not become running in expected time")
+			}, testutils.TestAPITimeout*45, testutils.TestRetryInterval).Should(BeTrue(), "Pod did not become running in expected time")
```
thom311, line 393: this commit fixes up an earlier commit of the PR. Please squash them.

internal/testutils/testcluster.go
```diff
@@ -21,3 +40,113 @@ func GetPod(c client.Client, name, namespace string) *corev1.Pod {
...
+		},
+		Spec: corev1.PodSpec{
+			NodeSelector: map[string]string{
+				"kubernetes.io/hostname": nodeHostname,
+			},
+			Containers: []corev1.Container{
+				{
+					Name:            "appcntr1",
+					Image:           "ghcr.io/ovn-kubernetes/kubernetes-traffic-flow-tests:latest",
+					Command:         []string{"/bin/sh", "-c", "sleep infinity"},
```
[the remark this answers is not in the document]
SalDaniele: > if you don't run the container in a separate pid namespace, then running the ENTRYPOINT (which uses tini) is not the right thing.

Having trouble parsing out what you are trying to say here....your point in general makes sense to me. I want trying to make the entrypoint explicit to the user, but if we rely on the knowledge that the kubernetes-traffic-flow-tests provides a reasonable default, I think that is fine.

The point about using tini to avoid zombie processes makes sense to me as well

pkgs/vars/vars.go
```diff
@@ -4,4 +4,6 @@ var (
 	Namespace = "openshift-dpu-operator"
 
 	DpuOperatorConfigName = "dpu-operator-config"
+
+	DefaultHostNADName = "default-sriov-net"
```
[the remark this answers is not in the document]
SalDaniele: This is actually a hardcoded NAD that is created by default by the dpu operator: https://github.com/openshift/dpu-operator/blob/main/internal/controller/bindata/networkfn-nad-host/00.networkfun-nad.yaml#L4

So I think it makes sense to have a variable in vars for both tests and core code to reference.

Makefile
SalDaniele: Yeah I thought about this as well, and wanted to discuss this with you. I wasn't sure how Marvell handles the secondary network traffic (i.e. traffic from workloads / network functions). I also wasn't sure if this was working yet in general.

In the case of the IPU both primary and secondary networking (from the perspective of Microshift) is handled on the same port.

If Marvell does something different we might need to refactor the lab to add an additional connection back to the provisioning host, so the provisioning host can server as an external client to validate secondary traffic, rather than re-using the same port as the primary as we are able to do in the case of the IPU.

If the same is true for Marvell, this will work, we would just need to update these environment variables at some point in the test to use the proper EXTERNAL_CLIENT_DEV for the server that Marvell runs on. In the past I think we have offloaded a lot of knowing which DPU something is running on to the spreadsheet, but I don't think this makes sense here. We could just have the Jenkins test that kicks this off set these environment variables explicitly.

WDYT?

SalDaniele: Will squash this with the previous commit

internal/testutils/testcluster.go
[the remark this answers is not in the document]
thom311: > I want trying to make the entrypoint explicit to the user

I think that makes some sense.

However, that leads to use `sleep infinity`, which is not a good PID1. And trying to run a more suitable PID1 in an arbitrary container, while assuming as little as possible about the container (e.g. whether `tini` is installed) is not going to work either.

At that point, I think it's reasonable to assume we know what container this is, what it has installed, what it's default ENTRYPOINT/CMD is.

SalDaniele:
@thom311 PTAL

PR #335, opened 2025-03-20
Fix incorrect vsi id extraction and duplicate p4 rule entry

arunvelayutham:
Address a possible incorrect VSI id usage scenario along with P4 rule add failure handle when a similar rule is already programmed or in place


SalDaniele:
@arunvelayutham please add `go mod tidy / go mod vendor` commit, looks like jobs are failing due to this

arunvelayutham:
Sorry missed to do "go mod vendor" before I pushed the changes while I was on travel. Just took care of it now.

saynb:
Replaced by https://github.com/openshift/dpu-operator/pull/339

PR #336, opened 2025-03-21
PV and PVC for vsp-p4

saynb:
* Adding RBAC waterfall for clusterroles, pv, pvc, roles

* Letting VSP have hostPID=true access. 
  * This is needed so that VSP can access the namespace of VSP-P4 even if VSP-P4 is created after VSP. Earlier, it only worked if VSP-P4 was created earlier

* Updating ipuplugin VSP refpoint 
  * Creating PV and PVC
  * Adding better pod wait and service IP resolution logic
  * Deleting all other VSP-P4 CRs except PV and PVC every time VSP is restarted


internal/daemon/plugin/bindata/vsp-ds/04.vsp_cluster_role.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: ClusterRole
+metadata:
+  name: vsp-cluster-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  - persistentvolumes
+  #resourceNames: vsp-p4-pv
```
SalDaniele, line 11: Left over comment?

saynb: Yes. Removed it

PR #338, opened 2025-03-24
Init Intel p4sdk Dockerfile

SalDaniele:
Add a dummy Dockerfile to unblock creating the requisite metadata to onboard the p4sdk container with ART


SalDaniele:
@bn222 @saynb PTAL

PR #345, opened 2025-03-26
Updating VSP so that PVC can be mounted. IIC443 IIC-421

saynb:
IIC-443 PVC
IIC-421 missing directory

Needs changes from https://github.com/bn222/cluster-deployment-automation/pull/326 in main before CI can be run

Edit: Above is merged. Needs a P4_URL from where the p4.tar.gz can be downloaded.


bn222:
using https://github.com/openshift/dpu-operator/pull/350 instead. New CDA with a fix is needed since OCP implementation changed.

/close

PR #347, opened 2025-03-27
[th/build-with-centos] honor BUILD_PATCH_DOCKERFILES=1 to build containers based on CentOSStream9

thom311:
With this change, we can `export BUILD_PATCH_DOCKERFILES=1` before building.

If so, we use a patched Dockerfiles, that use base images from CentOS stream 9.

The main purpose is to workaround https://issues.redhat.com/browse/ART-11976 .
This is only for development and quick testing. Usually, we still want to use the images from `registry.ci.openshift.org/`


Makefile
```diff
@@ -308,14 +308,44 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
 	done
 	@echo "Namespace 'openshift-dpu-operator' has been removed."
```
bn222, line 310: I like the idea of having a way to build with centosstream instead. the tmp files in-tree wont work on my setup. Can you use temp files like we already do in the makefile elsewhere? I will review the PR more deeply soon

Makefile
```diff
@@ -268,7 +274,6 @@ docker-build: test ## Build docker image with the manager.
 docker-push: ## Push docker image with the manager.
 	$(CONTAINER_TOOL) push ${IMG}
 
-GO_CONTAINER_CACHE = /tmp/dpu-operator-cache
```
bn222, line 271: By removing this path, it breaks on my setup. Can we keep the temp file in /tmp?

Makefile
thom311: could you elaborate why that does not work?

bn222: - The repository on the server is synchronized with my local directory. While it's building, it gets cleaned up
- In addition, in the CI, the whole directory gets removed every time we build, and this slows down because the cache isn't kept across builds

bn222:
Thank you for fixing the /tmp path. I have 2 comments remaining for this PR:

1. Can we have all the temp files be under that same path? That includes temp docker file as well.
2. Can we use go instead of .sh for our build system tools? For example, we already have: https://github.com/openshift/dpu-operator/blob/main/tools/incremental/incremental.go . It's not uncommon for projects to have a quite extensive build system that has logic in it and I see the need for it here too. For me, a shell script that encapsulates that logic is harder to maintain than go code.

thom311:
> Thank you for fixing the /tmp path. I have 2 comments remaining for this PR:
> 
>     1. Can we have all the temp files be under that same path? That includes temp docker file as well.

That would then be a future change, which might make sense. In general, in times of git this seems less of a concern (`git clean` works). What I find more concerning is to have files outside the working directory (e.g. `/tmp/cda-venv`). In any case, that seems unrelated to this PR.

>     2. Can we use go instead of .sh for our build system tools? For example, we already have: https://github.com/openshift/dpu-operator/blob/main/tools/incremental/incremental.go . It's not uncommon for projects to have a quite extensive build system that has logic in it and I see the need for it here too.

The project has various shells scripts (for good reasons). Also, `make` primarily runs shell scripts.

```
$ git grep -l '#!.*sh' -- ':(exclude)vendor/'
cmd/intelvsp/p4runtime-2023.11.0/copy_p4rt_python_deps.sh
cmd/intelvsp/p4sdk/entrypoint.sh
hack/both.sh
hack/deploy_fast.sh
hack/dpu_deploy.sh
hack/dpu_deploy_post.sh
hack/ipu_host_deploy.sh
hack/ipu_host_deploy_post.sh
hack/prepare-submodules.sh
hack/prepare-venv.sh
hack/traffic_flow_tests.sh
openshift/install-dpu.sh
scripts/check-gittree-for-diff.sh
scripts/patch-dockerfile.sh
```

> For me, a shell script that encapsulates that logic is harder to maintain than go code.

I disagree with that opinion.  Shell is a perfectly suitable tool for certain uses (like this one).

Why is there `hack/prepare-venv.sh` instead of `tools/prepare-venv/prepare-venv.go`? If this were a sensible thing to do, then was this not created that way? The answer is IMO, that shell is a better solution in this case.

bn222:
> more concerning is to have files outside the working directory

In-tree builds have downsides too. I think the balance might be to have way to tell it where to build. Thinking more about this, it seems that everything could go in a `BIN_DIR` which could be `bin/`. That way, we can do both in-repo builds and out-of-repo builds.

> I disagree with that opinion. Shell is a perfectly suitable tool for certain uses (like this one).

Let's have an off thread discussion about this.

> Why is there hack/prepare-venv.sh instead of tools/prepare-venv/prepare-venv.go? If this were a sensible thing to do, then why was it not written in go? The answer is IMO, that shell is a better solution in this case.

The reason is that we don't have a go wrapper around cda (and tft which was initially based on the same code base), which is something I think we should work on. That way, we can simplify managing the venv and replace it by a simple line in go.mod. Currently we don't have that because of time constraints. However, we should definitely discuss ideals vs. reality and then determine how close we want to be to ideals.

In any case, thanks for working on our build system.

PR #349, opened 2025-04-03
Marvell: Add Support for NetworkFunction in VSP

alkama-hasan:
This commit adds support for CreateNetworkFunction(), DeleteNetworkFunction() API to support traffic flow through NF in Marvell VSP

### DIFF
--- a/internal/daemon/vendor-specific-plugins/marvell/main.go
+++ b/internal/daemon/vendor-specific-plugins/marvell/main.go
@@ -44,6 +44,9 @@ const (
 	PFID           int    = 0
 	isDPDK         bool   = false
 	HostVFDeviceID string = "b903"
+	DpuRpmDeviceID string = "a063"
+	NfName         string = "mrvl-nf1"
+	isNf           bool   = false
 )
 
 // multiple dataplane can be added using mrvldp interface functions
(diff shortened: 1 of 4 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -44,6 +44,9 @@ const (
 	PFID           int    = 0
 	isDPDK         bool   = false
 	HostVFDeviceID string = "b903"
+	DpuRpmDeviceID string = "a063"
+	NfName         string = "mrvl-nf1"
```
wizhaoredhat, line 48: This feels a bit off, why do we hardcode this name?

alkama-hasan: Actually the intent here was to maintain all the port (vfs,inport,outport) associated with a particular Network Function. based on some nf identifiers (nf name here )
since right now we don't have flexibility/options to have it not a const  
because CBP calls are not specific to a pariticular nf neither CNF  (does not have any parameter related to nfs)  .
Ideally this identifiers should come from CBP and CNF let's say if we have multiple nfs and a CBP call  comes, how will vsp know this call is for which nf?
let me know your thoughts on this

PR #350, opened 2025-04-03
P4 update

bn222:


saynb:
Looks like the test has been stuck for 7 hours now. I saw this a couple of times earlier too. The timeout is set to 8 hours and it will die after that. Retest triggers don't help before the timeout either

saynb:
We can close this

PR #351, opened 2025-04-03
[th/cda-options] hack: pass $CDA_OPTIONS environment to cda call

thom311:
By default, our cda.py invocations don't enable debug logging. That makes debugging hard.

At least for manual usage, it will be useful if we are able to set `export CDA_OPTIONS='-v debug'` to override that setting.


saynb:
@thom311 @vrindle  Do you all know where the jobs are getting stuck on? Most jobs look like they are stuck and timing out at 8 hours. We are finding it hard to retrigger these jobs before timing out either

PR #353, opened 2025-04-07
Bump CDA

bn222:


Makefile
```diff
@@ -313,6 +313,17 @@ local-build: ## Build all container images necessary to run the whole operator
 	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.IntelVSP.rhel -t $(INTEL_VSP_IMAGE)
 	$(CONTAINER_TOOL) build -v $(GO_CONTAINER_CACHE):/go:z -f Dockerfile.networkResourcesInjector.rhel -t $(NETWORK_RESOURCES_INJECTOR_IMAGE)
 
+.PHONE: pre-pull-images
+pre-pull-images: tools
```
thom311, line 317: how does `pre-pull-images` achieve "to avoid race"?

What is the nature of this race, and how does this patch fix it?

bn222: We run make with `-j 8`. That means multiple targets run at the same time, including all the images. The build step pulls in the dependencies, and since multiple pulls of the same image happen concurrently, it hits a "can't reach image" error. 

To reduce the cache miss, we pull the images once before building.

Makefile
```diff
@@ -313,6 +313,17 @@ local-build: ## Build all container images necessary to run the whole operator
...
+	@if bin/base_images; then \
+		for image in $$(bin/base_images); do \
+			podman pull --platform=linux/amd64 $$image && \
+			podman pull --platform=linux/arm64 $$image; \
+		done \
+	else \
+		echo "Failed to get base images"; exit 1; \
+	fi
```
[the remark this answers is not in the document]
bn222: I would like to avoid logic in shell scripts. The for loop I've added is because we need something quick to fix the races, however I would like to avoid shell scripts going forward as we move towards a more complex build system.

Makefile
thom311: Do you mean `make -j 8 local-buildx`?

That target is not suitable for parallel execution, because dependencies like:
```
local-buildx: prepare-multi-arch go-cache local-buildx-manager local-buildx-daemon local-buildx-marvell-vsp local-buildx-intel-vsp local-buildx-network-resources-injector
```
will all build in parallel and in ill defined order. For example, `go-cache` might run after `local-buildx-manager`. What would need to happen, is that `local-buildx-manager` depends on `go-cache`, and not that `local-buildx` depends on `go-cache local-buildx-manager ` in parallel. That particular problem would be fixed by !347, but it seems there lurk many places where dependencies are not correctly expressed, and a `-j` build is likely to fail.

bn222: I don't see how a -j will still fail after this PR. Can you elaborate? My PR ensures that we pull the images before proceeding and then assume that no new image is available shortly after.

thom311: I was talking about the dependencies `local-buildx: go-cache local-buildx-manager`, which were wrong and which this commit also fixes. I missed that. Maybe the commit message should describe that this is fixed. Or maybe even have that fix in a separate commit, where the commit message is a place to discuss that fix.

This patch also fixes something what is on review for more than a week on #347...

bn222: Thanks for mentioning it. I pushed a change to make it more explicit. I will review !347 as soon as I can. First, we need to make sure our CI is passing.

e2e_test/e2e_test.go
```diff
@@ -350,8 +349,6 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 			// This should be the IP we expect to be able to reach from the DPU's secondary network
 			externalClientIp, err = getEnv("EXTERNAL_CLIENT_IP")
 			Expect(err).NotTo(HaveOccurred())
-			// This should be the dev on the external client that is connected to the DPU's secondary network
-			externalClientDev, err = getEnv("EXTERNAL_CLIENT_DEV")
```
thom311, line 354: it would be useful if the commit message would explain why this is done.

Is there a problem with the test? Obviously there is, but which is the problem?

Depending on that, it tells whether this test should ever be restored (in a similar form and how to changed it).

bn222: No idea what's wrong. I don't this this should have ever worked.

We've discussed this with Intel. They will revert the removal in a PR that fixes the behavior.

PR #359, opened 2025-04-11
[th/tft-any-cluster] make traffic_flow_tests more configurable to use with any cluster

thom311:
`make e2e-test` and `make traffic-flow-tests` call
`hack/traffic_flow_tests.sh`, which runs the tests. It uses a
configuration file and detects part of the configuration.

Most notably it assumes to find the KUBECONFIG files at a certain place
and find a worker node with a certain naming pattern in `oc get node`.

This is useful for when we run the script as part of the `e2e-test,
so that the cluster is setup in a certain way that works for the traffic
flow tests script.

However, it would be useful to be able to run the tests with any
cluster, without having to follow the naming choices.

As "hack/traffic_flow_tests.sh" is quite trivial, you could just fork
the script and modify it to your needs. But optimally, we want a way to
run the test from this repository (mostly) as they are, without forking
parts.

Make that possible by honoring the environment variables

  - `$TFT_KUBECONFIG`
  - `$TFT_KUBECONFIG_INFRA`
  - `$TFT_WORKER`

with their obvious(?) meaning. Set these to overwrite what is being
tested.

### DIFF
--- a/hack/traffic_flow_tests.sh
+++ b/hack/traffic_flow_tests.sh
@@ -1,22 +1,37 @@
 #!/usr/bin/env bash
 
+# The script detects the used parameters, on the assumption the cluster
+# was deployed by other scripts and via cluster-deployment-automation.
+#
+# You can however override this detection by setting the following
+# environment variables:
+# - TFT_KUBECONFIG (defaults to /root/kubeconfig.ocpcluster)
+# - TFT_KUBECONFIG_INFRA (defaults to /root/kubeconfig.microshift)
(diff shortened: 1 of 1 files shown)


hack/traffic_flow_tests.sh
```diff
@@ -27,7 +42,9 @@ temp_file=$(mktemp)
```
bn222, line 42: - We will replace this file by golang when we have CDA resume working
- Can you elaborate on why we need to support setting the kubeconfig path?

thom311: ok.

PR #360, opened 2025-04-11
introduce platform level abstraction

bn222:
Stepping stone towards adding DPU CR


bn222:
/test make-e2e-test

testing 223 server

internal/platform/vendordetector.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	vspPlugin, err := detector.VspPlugin(dpuMode, vspImages, client)
+	if err != nil {
+		return nil, errors.Errorf("Error encountered when deploying VspPlugin: %v", err)
+	}
+
+	return vspPlugin, nil
+}
+
+func (pi *PlatformInfo) Getvendorname() (string, error) {
```
wizhaoredhat, line 60: Function name should be GetVendorName following convention.

internal/platform/vendordetector.go
```diff
@@ -0,0 +1,193 @@
...
+				klog.Infof("Platform Host has %v's DPU detected", detector.GetVendorName())
+				return detector.GetVendorName(), nil
+			}
+		}
+	}
+
+	return "", errors.Errorf("No vendor found")
+}
+
+func (pi *PlatformInfo) GetPcieDevFilter() (string, string, string, error) {
```
wizhaoredhat, line 98: I think named return variables would be better here. Since there are 4 return values.

internal/platform/vendordetector.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, pci := range devices {
+		for _, detector := range pi.Detectors {
+			isDPU, err := detector.IsDPU(*pci)
+			if err != nil {
+				return "", "", "", err
+
+			}
+
+			if isDPU {
+				return pci.Vendor.ID, pci.Product.ID, pci.Address, nil
```
wizhaoredhat, line 112: The function name is confusing. Only if isDPU is true, we return something? When the function name GetPcieDevFilter does not mention it.

internal/platform/vendordetector.go
bn222: This is still temporary. Just a function move. Next patch with address more of this

internal/platform/vendordetector.go
bn222: Agree, this is just a move still. Next PR will rework vendor detection completely

internal/platform/vendordetector.go
bn222: Still only a move. Will address during detection rework

internal/platform/vendordetector.go
```diff
@@ -0,0 +1,193 @@
...
+	klog.Infof("Detecting  Platform is DPU or not")
+	for _, detector := range pi.Detectors {
+		isDpu, err := detector.IsDpuPlatform(pi.p)
+		if err != nil {
+			return "", err
+
+		}
+
+		if isDpu {
+			klog.Infof("Platform is a %v's DPU detected", detector.GetVendorName())
```
wizhaoredhat, line 70: I think since we have this print anyways to give Vendor Name and description.

internal/platform/vendordetector.go
```diff
@@ -0,0 +1,193 @@
...
+	return vspPlugin, nil
+}
+
+func (pi *PlatformInfo) Getvendorname() (string, error) {
+	klog.Infof("Detecting  Platform is DPU or not")
+	for _, detector := range pi.Detectors {
+		isDpu, err := detector.IsDpuPlatform(pi.p)
+		if err != nil {
+			return "", err
+
```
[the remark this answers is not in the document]
bn222: Ok, I'll fix it in the next PR that follows this one

internal/platform/vendordetector.go
wizhaoredhat: Just a comment for later.

internal/platform/vendordetector.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return "", errors.Errorf("Failed to get PCI devices: %v", err)
+	}
+
+	for _, pci := range devices {
+		for _, detector := range pi.Detectors {
+			isDPU, err := detector.IsDPU(*pci)
+			if err != nil {
+				return "", err
+
```
wizhaoredhat, line 85: NIT: Extra new line.

PR #366, opened 2025-04-16
Reintroduce pod/nf to pod on remote node testcase on dpu-operator

arunvelayutham:
This reverts commit 1cef24934aa621e518258ff4840cc667b3583906.

### DIFF
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -13,6 +13,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -89,6 +90,35 @@ func setupExternalClient(externalClientIp, externalClientDev, workloadSubnet str
 			fmt.Printf("Warning: Failed to remove route: %v, output: %s\n", err, output)
 		}
 	})
+}
+
+func createAndWaitForNFPod(ctx context.Context, dpuSideClient client.Client, sfc client.Object) (*corev1.Pod, error) {
+	// Ensure no stale NF pod is hanging around
+	staleGone := func() (bool, error) {
+		return testutils.GetPod(dpuSideClient, nfName, vars.Namespace) == nil, nil
+	}
+	if err := wait.PollImmediate(interval, timeout, staleGone); err != nil {
+		return nil, fmt.Errorf("old NF pod %q still present after %s: %w", nfName, timeout, err)
+	}
+	// Create NF Pod using ServiceFunctionChain custom resource
+	if err := dpuSideClient.Create(ctx, sfc); err != nil {
+		return nil, fmt.Errorf("creating NF Pod: %w", err)
+	}
+	// Poll until the NF pod is Running with the expected image
+	var nfPod *corev1.Pod
+	ready := func() (bool, error) {
+		nfPod = testutils.GetPod(dpuSideClient, nfName, vars.Namespace)
+		if nfPod == nil {
+			return false, nil // not created yet
+		}
+		isRightImage := nfPod.Spec.Containers[0].Image == nfImage
+		isRunning := nfPod.Status.Phase == corev1.PodRunning
+		return isRightImage && isRunning, nil
+	}
+	if err := wait.PollImmediate(interval, timeout, ready); err != nil {
+		return nil, fmt.Errorf("NF pod %q failed to reach Running within %s: %w", nfName, timeout, err)
+	}
+	return nfPod, nil
 }
 
 func TestControllers(t *testing.T) {
@@ -308,6 +338,7 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 			nfIngressIp                string
 			nfEgressIp                 string
 			externalClientIp           string
+			externalClientDev          string
 			externalSubnet             string
 			skipNetworkFunctionTesting = false
 			sfc                        *configv1.ServiceFunctionChain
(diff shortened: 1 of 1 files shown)


SalDaniele:
/retest

Seems to be failing because of an issue with the Network cluster operator.

Error when the daeomn is starting (host side only)

```
50
  Normal   Scheduled  137m                    default-scheduler  Successfully assigned openshift-dpu-operator/dpu-daemon-zbxp2 to worker-215
  Normal   Pulled     137m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 1.514s (1.514s including waiting). Image size: 535023161 bytes.
  Normal   Pulled     137m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 61ms (61ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     137m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 67ms (67ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     136m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 62ms (62ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     136m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 68ms (68ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     136m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 65ms (65ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     136m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 69ms (69ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     135m (x4 over 137m)     kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 66ms (66ms including waiting). Image size: 535023161 bytes.
  Normal   Pulled     135m                    kubelet            Successfully pulled image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev" in 63ms (63ms including waiting). Image size: 535023161 bytes.
  Normal   Pulling    2m46s (x618 over 137m)  kubelet            Pulling image "wsfd-advnetlab223.anl.eng.bos2.dc.redhat.com:5000/dpu-daemon:dev"
  Warning  Failed     2m46s (x618 over 137m)  kubelet            Error: container create failed: cannot stat `/usr/lib/opt/cni`: No such file or directory      
```

bn222:
@arunvelayutham this needs a rebase

e2e_test/e2e_test.go
```diff
@@ -308,6 +308,7 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 			nfIngressIp                string
 			nfEgressIp                 string
 			externalClientIp           string
+			externalClientDev          string
```
thom311, line 311: @arunvelayutham could you please explain (in the commit message) why things are done?

The commit message says hat this is a revert of 1cef249. But the original commit 1cef249 does not explain why the test was removed in the first place. And this PR does not explain, why the removal is now reverted. What changed, so that it's OK to bring back this test?

e2e_test/e2e_test.go
```diff
@@ -450,6 +453,40 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 				pingTest(dpuClientSet, dpuRestConfig, nfPod, pod1_ip, nfPod.Name, pod1.Name)
 				pingTest(dpuClientSet, dpuRestConfig, nfPod, pod2_ip, nfPod.Name, pod2.Name)
 			})
+			g.It("Should support Network-function -> external", func() {
+				if skipNetworkFunctionTesting {
+					g.Skip("Skipping Network Function Testing")
+				}
+
+				fmt.Printf("Assigning NF ingress port IP %s\n", nfIngressIp)
+				_, err := testutils.ExecInPod(dpuClientSet, dpuRestConfig, nfPod, fmt.Sprintf("ip addr add %s dev net2", nfIngressIp+"/24"))
```
wizhaoredhat, line 462: Could you explain why you need to run these ip route commands?

e2e_test/e2e_test.go
thom311: Hi, you now changed the commit message
```diff
-    Revert "remove pod/nf to pod on remote node test"
-    
-    This reverts commit 1cef24934aa621e518258ff4840cc667b3583906.
+    Reintroduce Pod to Pod on remote node connectivity test
+    
+    Reintroduce Pod/NF to Pod on remote node connectivity test on
+    dpu-operator CDA framework.
```
The previous commit message quite aptly described "what" is happening. The new message still only talks about "what", but is almost worse, because it no longer describes that this resurrects/reverts and earlier commit (and which exact commit).

What however is still missing, is *why* any of this is done. E.g.
- why was the test dropped in 1cef249 (which should have been described in 1cef249, but wasn't)
- what changed, so that it can be brought back.

For example, something like "In 1cef249 the test was removed because of bug XYZ (more references) and issue UVW, which caused the bug to fail. These problems are now fixed by fix ABC (more references) and the test can be restored."

And also mention IIC-582 (if that issue is indeed about this PR), because that has potentially further information.

Anyway, I don't want to hamper this too much. Sorry about that. So if you think this is unnecessary, then let's drop it.

e2e_test/e2e_test.go
wizhaoredhat: You mean that the operator does not write IPAM into the config JSON, is that correct?

e2e_test/e2e_test.go
arunvelayutham: @thom311 @wizhaoredhat I created a new PR with just the revert. As you mentioned, I originally intent to keep it 2 commits (1 to revert and 1 to extend the test scope). But I accidently did an amend on the revert commit itself. To keep it neat, I have raised a new PR https://github.com/openshift/dpu-operator/pull/480

Let me close PR-366 against this new PR-480 if that is good with you

thom311: sounds good to me.

The commit message from #480 also is good. Thank you!!

PR #369, opened 2025-04-21
Intel IPU SDK2.0 based FXP Linux Networking package upgrade

arunvelayutham:
fxp-net_linux-networking package compiled with SDK2.0 based P4 compiler version 3.0.70.134. This is required to upgrade Redhat image to IPU SDK2.0 release.

PR #372, opened 2025-04-28
Sdk2.0

bn222:


bn222:
/test make-e2e-test

new kind of flake, maybe related to 2.0

bn222:
/test make-e2e-test

p4 tar gz was not on the webserver

saynb:
Looks like a false pass. It ran for 46 seconds and passed. Will run again

/test make-e2e-test

bn222:
/test make-e2e-test

test ran successfully, need to pick up the results

e2e_test/e2e_test.go
```diff
@@ -309,6 +309,7 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 			nfIngressIp                string
 			nfEgressIp                 string
 			externalClientIp           string
+			externalClientDev          string
```
thom311: Please explain in the commit message shortly why this test was removed in the first place. Well, optimally, you refer to the commit that removed it, and the reason for the removal is there (actually, there isn't a reason given there).

But more importantly. Now that the removal is reverted, please explain why this is undone.

This can be as simple as "This test was known broken and got removed earlier. In the meantime it was fixed, so restore the test". Optimally, explain briefly what the issue was (or which fix was done to make the test work).

But don't write it here. Please write it in the commit message.

e2e_test/e2e_test.go
```diff
@@ -309,7 +309,6 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 			nfIngressIp                string
 			nfEgressIp                 string
 			externalClientIp           string
-			externalClientDev          string
```
thom311: Why does the same PR re-introduce a previously removed test, and then deletes it again.

Probably, don't do that.

If you really want to do this, then please explain in the commit message why this is done.

thom311 (commented):
this PR contains part of https://github.com/openshift/dpu-operator/pull/369 .

Preferably get 369 in first (and rebase this PR on top of that).
Alternatively, close 369 and cleanup the history of this PR (e.g. not merging main into the PR).

PR #384, opened 2025-05-22
[th/redeploy-target] add "redeploy" and "run-e2e-test-suite" targets

thom311:
- taskfile: add "run-e2e-test-suite" taskfile target
- taskfile: add "redeploy" and "run-e2e-test-suite" taskfile targets
- taskfile: fix URL for network-resources-injector-image:dev image

### DIFF
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -137,15 +137,19 @@ tasks:
       - task: deploy-clusters
       - task: fast-e2e-test
 
-  fast-e2e-test:
-    deps:
-      - task: ginkgo
-      - task: envtest
+  redeploy:
     cmds:
       - task: build-image-all
       - task: undeploy
       - task: deploy
       - sh hack/setup.sh
+
+  fast-e2e-test:
+    deps:
+      - task: ginkgo
+      - task: envtest
+    cmds:
+      - task: redeploy
       - >
         FAST_TEST=true
         NF_INGRESS_IP=10.20.30.2
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -144,12 +144,11 @@ tasks:
       - task: deploy
       - sh hack/setup.sh
 
-  fast-e2e-test:
+  run-e2e-test-suite:
     deps:
       - task: ginkgo
       - task: envtest
     cmds:
-      - task: redeploy
       - >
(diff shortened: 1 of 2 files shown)


taskfile.yaml
```diff
@@ -144,17 +144,21 @@ tasks:
       - task: deploy
       - sh hack/setup.sh
 
-  fast-e2e-test:
+  run-e2e-test-suite:
```
bn222, line 147: when would you run the test without deploying the operator? You will always need to use the `fast-e2e-test` target. It doesn't make sense to have a separate target for just running the test that is doesn't have `internal: true`

thom311: - to deploy the operator once, and run the test multiple times
- to run the test after the operator was already deployed earlier

bn222: why would you want to run the same tests multiple times? I understand that we need to run it, but exposing this complicates things.

thom311: For testing and debugging the test.

bn222: iiuc, you're saying you would make a change to the test while keeping the code the same and want to build and test the test only?

taskfile.yaml
```diff
@@ -110,7 +110,7 @@ tasks:
       - task: push-image-helper
         vars:
           SOURCE: 'localhost/network-resources-injector:dev'
-          IMAGE: '{{.REGISTRY}}:5000/network-resources-injector:dev'
+          IMAGE: '{{.REGISTRY}}:5000/network-resources-injector-image:dev'
```
bn222, line 113: This was a bug? Why is it different from the other images?

taskfile.yaml
```diff
@@ -137,20 +137,28 @@ tasks:
       - task: deploy-clusters
       - task: fast-e2e-test
 
-  fast-e2e-test:
-    deps:
-      - task: ginkgo
-      - task: envtest
+  redeploy:
```
bn222, line 140: Update the readme to refer to this too

taskfile.yaml
bn222: Good catch, but we should fix it so that it's uniform at this level. I'll push a PR for the fix

PR #397, opened 2025-05-29
Pull remote image into local registry

bn222:
Instead of using a timeout to bring up a pod that depends on a resource on the internet, pull in the image to the local registry and only capture the timeout of starting the pod using that local image which is more deterministic. This removes a flake in the CI when the remote server was acting slower than the hard coded timeout of 2 minutes.

### DIFF
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -118,6 +118,7 @@ tasks:
     cmds:
       - >
         FAST_TEST=true
+        REGISTRY={{.REGISTRY}}
         NF_INGRESS_IP=10.20.30.2
         EXTERNAL_CLIENT_DEV=eno12409
         EXTERNAL_CLIENT_IP=10.20.30.100


thom311:
This seems too much, for what a test should do. Also, now the test requires that somebody first starts a container registry and sets `REGISTRY` environment variable. It increases the complexity for just trying to run the tests.

If you want to ensure that the DPU has the image ready, maybe something like `oc debug $NODE /host/usr/bin/crictl --runtime-endpoint=unix:///host/run/crio/crio.sock pull --timeout 60 $IMAGE` would be easier. 

Or maybe better: the test should be robust against the pod taking a while to start.

> Or maybe better: the test should be robust against the pod taking a while to start.

The timeout really only seems to be useful to not hang too long in case where the test is clearly failing. That means, there is no major problem to set a high timeout, that covers all slow runs.

wizhaoredhat:
I agree with Thomas. It seems like having a local registry as a pre-req is too much of an ask . Is polling for pod ready status not good enough?

wizhaoredhat:
Hmmm is it possible to add, if internal registry not found, fall back to internet?

wizhaoredhat:
This is ok for me. @thom311 please share what your thoughts are. Thanks!

wizhaoredhat:
/hold 
@bn222 Could you please add USE_REMOTE_IMAGES flag?

bn222:
yes, currently working on other PRs first because we have merged a fix for IPU specifics that is causing some CI flakes. I will add the agreed upon flag to this PR when I rework it. Thank you for adding the `/hold` to avoid accidentally merging it

PR #398, opened 2025-05-29
CI for removing obsolete redfish workaround

swtaylor:


bn222:
Thank you for the patch. It passes, so I've merged your change into CDA and have now pushed a new PR to update to the in-stream head of CDA: https://github.com/openshift/dpu-operator/pull/399

PR #405, opened 2025-06-03
[th/e2e-test-skip-nf] e2e_test: skip pod2pod test with NF for SKIP_NF_TESTING=1

thom311:
Marvell does not yet pass this test. The caller (Jenkins) will set SKIP_NF_TESTING=1 when running on Marvell, to get a pass on the overall suite -- with the exeception of the tests that are known to fail and which are opted out via SKIP_NF_TESTING.


bn222:
I understand what and why you're trying to do this and the approach is sound until we have the DPU CRs which we will use to skip the tests for Marvell unless the support for the NFs gets merged. Can you double check the code though?

thom311:
> Can you double check the code though?

What should I double check?

bn222:
I see that you're using `g.Skip`. Disregard my previous comment.

PR #411, opened 2025-06-05
Fix support for other registries

bn222:


bn222:
manually ran fast-e2e-test to go over the final bit. It passed. This failed on the first run due to a known flake.

PR #416, opened 2025-06-06
Intel VSP plugin version update

arunvelayutham:
This update covers restoration of primiary network path upon vsp-plugin delete triggered by dpu-operator clean-up


bn222:
@arunvelayutham please rebase

PR #422, opened 2025-06-10
CI for redfish install simplification

swtaylor:


bn222:
let's collect >2 green runs.

PR #427, opened 2025-06-17
[th/e2e-test-sfcs] e2e_test: add test creating multiple SFCs in parallel

thom311:
Create a bunch of SFCs. As the "openshift.io/dpu" resource is limited, they probably cannot start all right away. However, we are also deleting the SFCs after a short while, so eventually all SFCs should have their pods running and quit.

https://issues.redhat.com/browse/IIC-685

### DIFF
--- a/e2e_test/e2e_test.go
+++ b/e2e_test/e2e_test.go
@@ -435,6 +435,7 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 				nfPod = testutils.EventuallyPodIsRunning(dpuSideClient, nfName, vars.Namespace, timeout_sfc_pod_running, interval)
 
 				Expect(nfPod.Spec.Containers[0].Image).To(Equal(imageRef), "Pod should have expected image")
+				Expect(testutils.PodGetDpuResourceRequests(nfPod)).To(Equal(2))
 
 				fmt.Println("Nf pod successfully created")
 			})
--- a/internal/testutils/testcluster.go
+++ b/internal/testutils/testcluster.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift/dpu-operator/pkgs/vars"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -36,6 +37,18 @@ type NetworkStatus struct {
 	IPs       []string `json:"ips"`
 	Mac       string   `json:"mac"`
 	DNS       struct{} `json:"dns"`
+}
+
+func PodGetDpuResourceRequests(pod *corev1.Pod) int {
+	total := resource.MustParse("0")
+
+	for _, c := range pod.Spec.Containers {
+		if qty, ok := c.Resources.Requests["openshift.io/dpu"]; ok {
+			total.Add(qty)
+		}
+	}
+
+	return int(total.Value())
 }
 
 func GetPod(c client.Client, name string, namespace string) *corev1.Pod {


e2e_test/e2e_test.go
```diff
@@ -495,6 +620,17 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 					return err != nil && errors.IsNotFound(err)
 				}, testutils.TestAPITimeout*30*2, testutils.TestRetryInterval).Should(BeTrue(), "Pod %s was not fully deleted in time", podName)
 			}
+
+			for i := 0; i < sfcNum; i++ {
+				namespace, sfcName1, nfName1 := sfcGetName(i)
+				dpuSideClient.Delete(context.TODO(), sfcCreate(namespace, sfcName1, nfName1))
```
bn222, line 626: Deleting the SFC should cause the pods to be eventually deleted. There is no need to manually delete the pod

e2e_test/e2e_test.go
```diff
@@ -465,6 +477,119 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+
+					for _, state := range states {
+						if !state.ts_delete_at.IsZero() && now.After(state.ts_delete_at) {
+							state.ts_delete_at = time.Time{}
+							state.ts_wait_for_pod_deleted = now.Add(60 * time.Second)
+
+							fmt.Printf("SFC pod (%s) %s,%s: deleting SFC\n", vars.Namespace, state.sfcName, state.nfName)
+							err2 := dpuSideClient.Delete(context.TODO(), state.sfc)
+							Expect(err2).NotTo(HaveOccurred())
+							continue outer
```
bn222, line 566: Let's not use gotos.
Please simplify the patch to:
// should have no SFCs and no pods that would be created
// create 5 SFCs
// eventually should have 5 pods in ready state
// eventually should have the right amount of resources
// delete 5 SFCs
// eventually should have no pods
// eventually should have the right amount of resources

All of that done under "Should support creating and running multiple SFCs".

Think BDD, not imperative style: Any test can run at any time in any order. We want to work towards removing `g.Ordered`, not towards making the code depend on that more.

thom311: > should have no SFCs and no pods that would be created

Do you mean, there should be a check for the precondition of not having such SFCs? Creating an SFC will fail if an SFC of the same name already exists, so that is covered.

> // create 5 SFCs

Sure they could also be created all at once. The test chooses to create them slower, as SFC pods are starting up. The loop already initially creates more SFCs that are running, so the case of creating many SFCs that cannot all run is covered. Additionally, it creates more SFCs later, to see whether those also end up running. This is to cover a slightly different scenario.

> // eventually should have the right amount of resources

You mean, the test should check how many `openshift.io/dpu` resources are (in total and allocated)? OK. Why? What exactly should it check? "The right amount" depends on the system and not something that necessarily can be asserted against.

> // eventually should have 5 pods in ready state

5 pods cannot be in ready state, because likely there are not enough `openshift.io/dpu` resources.

bn222: > Do you mean, there should be a check for the precondition of not having such SFCs? Creating an SFC will fail if an SFC of the same name already exists, so that is covered.

If there is no pre-check, how can we tell that it failed because the test ran on an unclean env?

> Sure they could also be created all at once

If we're testing sequential creation, can we still simplify the code? Loop around creating + wait/check + delete + wait/check

> You mean, the test should check how many openshift.io/dpu resources are
Yes, because based on what Intel was saying, they were expressing that the number of resources reported here were not correct. 

> 5 pods cannot be in ready state

Disregard that comment since we're testing sequential creation.

```
g.Context("Multiple ServiceFunctionChains", g.Ordered, func() {
			g.It("Should support creating and running multiple SFCs", func() {
```

Clarify this please so that it's obvious that we're doing sequential creation. Today it sounds like we want to test having multiple of them at the same time.

thom311: > > Do you mean, there should be a check for the precondition of not having such SFCs? Creating an SFC will fail if an SFC of the same name already exists, so that is covered.
>
> If there is no pre-check, how can we tell that it failed because the test ran on an unclean env?

The code will try to create the SFC, which fails with AlreadyExists (which will fail the assertion).

Also, another test that you wrote also doesn't check that ([here](https://github.com/openshift/dpu-operator/blob/9a7288021b33bebaaebc614ff8ff0a5f358ea436/e2e_test/e2e_test.go#L101)). Or [here](https://github.com/openshift/dpu-operator/blob/d1ec588f8960ea2017f77545e58f19320cc8ca38/e2e_test/e2e_test.go#L337) (not written by you). The latter fails frequently, if I kill a test and cleanup is skipped, then the subsequent run will fail due to that left-over (but that failure will then run the cleanup, and the 3 run will again show the original problem).

We can always add more (and redundant) assertions. I am generally in favor of asserting a lot. It does not seem very useful in this case.

> If we're testing sequential creation, can we still simplify the code?

I reworked the code. Now all SFCs are created upfront. Better?

thom311: > The hand-rolled loops for checking pod status (ts_wait_for_pod) and deletion (ts_wait_for_pod_deleted) are verbose and reinvent the wheel.

What is your concrete suggestion?

bn222: > Gomega provides powerful asynchronous matchers, Eventually and Consistently, specifically for this kind of polling

Concrete suggestions: Use `Eventually` (and `Consistently`) where possible.

e2e_test/e2e_test.go
```diff
@@ -465,6 +477,108 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+
+					err1 := dpuSideClient.Create(context.TODO(), state.sfc)
+					Expect(err1).NotTo(HaveOccurred())
+
+					now := time.Now()
+					state.ts_created = now
+					state.ts_wait_for_pod = now.Add(300 * time.Second)
+				}
+
+				for !allDone(testStates) {
```
bn222, line 530: Use Eventually here

e2e_test/e2e_test.go
```diff
@@ -465,6 +477,108 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+							}
+							if pod == nil || pod.Status.Phase != corev1.PodRunning {
+								if testutils.HasTimeout(state.ts_wait_for_pod) {
+									g.Fail(fmt.Sprintf("SFC Pod (%s) %s,%s: is not running after timeout", vars.Namespace, state.sfcName, state.nfName))
+								}
+							} else {
+								wait_time := rand.Intn(7)
+								state.ts_wait_for_pod = time.Time{}
+								state.ts_delete_at = time.Now().Add(time.Duration(wait_time) * time.Second)
+								fmt.Printf("SFC pod (%s) %s,%s: pod is now running (stop in %d seconds)\n", vars.Namespace, state.sfcName, state.nfName, wait_time)
```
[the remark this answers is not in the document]
bn222: Using Eventually constructs, we can block at any time until the changes are propagated in the system. The test should test precisely one thing otherwise it becomes clouded and more difficult debug. We can add another test that waits a random amount of time, but let's keep the current test focused on what the It(...) block says it will test.

e2e_test/e2e_test.go
```diff
@@ -465,6 +477,105 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+					}
+
+					for _, state := range testStates {
+						if !state.done {
+							return true
+						}
+					}
+					return false
+				},
+					1000*time.Second,
```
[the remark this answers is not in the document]
thom311: do you mean `testutils.TestAPITimeout`?

You cannot have one timeout and use it for everything. This timeout is already in use (roughly the time it takes to create a Pod). Here, we create several SFCs, wait for them to run and quit. So this is an entirely (distinct) timeout.

Various places don't use `testutils.TestAPITimeout` directly. They will do instead `testutils.TestAPITimeout*30*2`.  Which just hides the actual value. Also, it's not clear why that operation is expected to take 60 times the base-unit. We already have a base unit that everybody understands: `time.Second`.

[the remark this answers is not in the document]
thom311: First of all, this timeout here is never to be reached because they code inside the test will timeout before that happens (`g.Fail()`). So this is just a timeout of very last resorts. I mean, you can argue that no timeouts (that we assert against) are supposed to be reached in a test. But here, this timeout cannot be reached due to the code of the test itself (which is even stronger than saying "this timeout is not reached because the code we test is fine").

Also, all timeouts used in this patch are timeouts of last resort. They mainly exist to not hang the test forever and fail with a slightly better error. Whenever they are reached, the test will fail. If you tomorrow make the test pass faster, you don't need to adjust any of those timeouts. This is not one of those timeouts where we wait for something to always happen (e.g. "ensure that the pod runs at least X amount of time"). Such latter kind you can tweak to be as short as possible to make the test complete faster.

Also, if you tomorrow improve something to make a certain aspect faster, then you might decrease `testutils.TestAPITimeout`. But that only works, if `testutils.TestAPITimeout` (or `testutils.TestAPITimeout * factor`) is used for things that are related to the thing you are going to improve. As said, the usage here is quite unrelated to other uses that also use `testutils.TestAPITimeout`. There is no other test that would want a timeout value that is (linearly or otherwise) related to the timeout value of this this. Hence, it doesn't use a shared base-unit like `testutils.TestAPITimeout` that you can adjust for all tests at once.

e2e_test/e2e_test.go
```diff
@@ -35,13 +36,65 @@ var (
...
+}
+
+func sfcNew_i(i int) *configv1.ServiceFunctionChain {
+	namespace, sfcName, nfName := sfcGetName_i(i)
+	return testutils.SfcNew(namespace, sfcName, nfName, nfImage)
+}
+
+func sfcFindRunningPod(c client.Client, sfcNum int) int {
+
+	found_idx := -1
```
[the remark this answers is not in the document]
thom311: return the sfc from where? What does that simplify?

The index is needed, so we know which `SFC${i}` we need to delete.

The kubernetes scheduler probably does not guarantee, that always the oldest is started first. In any case, the code shouldn't rely on that.

e2e_test/e2e_test.go
```diff
@@ -465,6 +509,49 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
 			})
 		})
+
+		g.Context("With more ServiceFunctionChains than resources", g.Ordered, func() {
+			g.It("Should have all SFC Pods eventually running", func() {
+
+				sfcNum := (getMaxDpuResources(dpuSideClient) / 2) + 2
+				if sfcNum > sfcNumMax {
+					sfcNum = sfcNumMax
+				}
```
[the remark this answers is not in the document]
thom311: This limit is to delete the potentially creates (left-over) SFCs from the *previous* run.

e2e_test/e2e_test.go
```diff
@@ -465,6 +507,50 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+				for i := 0; i < sfcNum; i++ {
+					sfc := sfcNew(i)
+					testutils.SfcCreate(dpuSideClient, sfc)
+
+					Eventually(func() bool {
+						return testutils.GetPod(dpuSideClient, sfc.Spec.NetworkFunctions[0].Name, sfc.ObjectMeta.Namespace) != nil
+					}, timeout, interval).Should(BeTrue())
+				}
+
+				for i := 0; i < sfcNum; i++ {
```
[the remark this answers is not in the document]
thom311: Why? The SFCs are now all created upfront. Their Pods will be created right away, and all be in PENDING.

e2e_test/e2e_test.go
```diff
@@ -465,6 +509,49 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+				}
+
+				g.By(fmt.Sprintf("Creating %d SFCs", sfcNum))
+				for i := 0; i < sfcNum; i++ {
+					sfc := sfcNew_i(i)
+					testutils.SfcCreate(dpuSideClient, sfc)
+
+					Eventually(func() bool {
+						return testutils.GetPod(dpuSideClient, sfc.Spec.NetworkFunctions[0].Name, sfc.ObjectMeta.Namespace) != nil
+					}, timeout, interval).Should(BeTrue())
```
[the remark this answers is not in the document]
thom311: at this point, clearly not all pods can become running. Some will stay in PENDING. So what's the point of doing an `Eventually(PodIsRunning())`.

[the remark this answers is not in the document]
bn222: You don't know which one is not ready and you need to find it. To simplify this, if we can create n SFCs, create n first, wait for them to be ready, then create the (n+1)st which will be in pending state.

e2e_test/e2e_test.go
```diff
@@ -313,20 +365,7 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
-			Spec: configv1.ServiceFunctionChainSpec{
-				NetworkFunctions: []configv1.NetworkFunction{
-					{
-						Name:  nfName,
-						Image: nfImage,
-					},
-				},
-			},
-		}
+		sfc := testutils.SfcNew(vars.Namespace, sfcName, nfName, nfName)
```
bn222, line 368: nfName twice?

bn222:
We still have 12 as a hardcoded number of NFs tying the test to implementation detail (moving from black box to white box testing). If we calculate how many NFs we should create, and then create however many would the test less brittle. The test is still not entirely clear to me. If we could allocate 5 SFCs on a system, and we have < 5 allocated, the code will pass, but if we allocated 5 it will fail, and it will fail in a hard-to-debug way.

The intent of the test is still unclear. Based on discussions, the current understanding is that if a pod of a sfc goes into pending, it cannot go out of it and the test should disprove exactly that.

@wizhaoredhat please chime in on the review.

e2e_test/e2e_test.go
```diff
@@ -468,6 +507,50 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+					testutils.SfcCreate(dpuSideClient, sfc)
+
+					Eventually(func() bool {
+						return testutils.GetPod(dpuSideClient, sfc.Spec.NetworkFunctions[0].Name, sfc.ObjectMeta.Namespace) != nil
+					}, timeout, interval).Should(BeTrue())
+				}
+
+				for i := 0; i < sfcNum; i++ {
+					g.By(fmt.Sprintf("Find a running SFC Pod (%d of %d)", i, sfcNum))
+					sfc := sfcFindRunningSfc(dpuSideClient, sfcNum)
```
wizhaoredhat, line 535: This is flaky because above when you create the SFC, the scheduler doesn't necessarily follow FIFO ordering.

e2e_test/e2e_test.go
```diff
@@ -468,6 +507,50 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
 				}, timeout, interval).Should(BeTrue())
 			})
 		})
+
+		g.Context("With more ServiceFunctionChains than resources", g.Ordered, func() {
+			g.It("Should have all SFC Pods eventually running", func() {
```
wizhaoredhat, line 513: If there is more SFCs then resources then it is not expected that SFC Pods would be eventually running. A subset would be running but the remainder would be in pending state.

e2e_test/e2e_test.go
```diff
@@ -468,6 +507,50 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+					sfcNum = sfcNumMax
+				}
+
+				g.By(fmt.Sprintf("Creating %d SFCs to allocate all DPU side resources", sfcNum))
+				for i := 0; i < sfcNum; i++ {
+					sfc := sfcNew(i)
+					testutils.SfcCreate(dpuSideClient, sfc)
+
+					Eventually(func() bool {
+						return testutils.GetPod(dpuSideClient, sfc.Spec.NetworkFunctions[0].Name, sfc.ObjectMeta.Namespace) != nil
```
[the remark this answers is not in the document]
bn222: The code is not waiting for it to be running. I proposed to have a creation loop of a number of SFCs that will all be in running state (block and wait on those before proceeding) before moving on to creating 1 more that will be in pending state. What do you think of that approach?

internal/testutils/testcluster.go
```diff
@@ -38,6 +38,28 @@ type NetworkStatus struct {
...
+		if qty, ok := c.Resources.Requests["openshift.io/dpu"]; ok {
+			total.Add(qty)
+		}
+	}
+
+	// For init containers we take the maximum.
+	for _, ic := range pod.Spec.InitContainers {
+		if qty, ok := ic.Resources.Requests["openshift.io/dpu"]; ok {
+			if qty.Cmp(total) > 0 {
+				total = qty
```
wizhaoredhat, line 55: I don't really understand this... Can Init containers have separate resource requests?

e2e_test/e2e_test.go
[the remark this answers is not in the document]
wizhaoredhat: I think the purpose of the E2E test is to try to catch the issue that we have with the IPU SFC cleanup being broken right? If so, maybe doing it sequentially will be more deterministic and we can see if it fails. If it doesn't fail we can inject randomness (by not waiting).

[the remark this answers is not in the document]
bn222: I agree with your proposal. It's much easier to understand when sequential code fails. Would it be more clear if we do:

1. Ensure no SFCs are running
2. Calculate the total number of SFCs that can be running simultaneously. Call that n.
3. Create n SFCs
4. Wait for each of those to be in running state
5. Create one more
6. Ensure that that one is in pending state
7. Delete the first running SFC
8. Ensure that the last pending SFC moves to running state
9. Remove all SFCs

If that's still too complex, we can simplify further

e2e_test/e2e_test.go
thom311: `sfcFindRunningSfc()` returns any running SFC that it finds. It does not rely on any particular ordering in which the scheduler starts the pods.

e2e_test/e2e_test.go
thom311: "eventually" indicates that they all run before the end of the test. It does not mean, that they all run at the same time, which they cannot and do not.

But sure. What would be a better name?

thom311: reworded

e2e_test/e2e_test.go
```diff
@@ -460,6 +503,84 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+				sfcLs := testutils.SfcList(dpuSideClient, vars.Namespace)
+				Expect(sfcLs.Items).To(BeEmpty())
+			})
+
+			var sfcNum int
+
+			g.It("Should initially detect number of DPU resources", func() {
+				dpuResNum := getMaxDpuResources(dpuSideClient)
+				Expect(dpuResNum).To(BeNumerically(">=", 2))
+				sfcNum = (dpuResNum / 2) + 1
```
bn222, line 518: It would be easier to read if sfcNum = dpuResNum/2, and the loop below just loops until sfcNum followed by a check if all of them are running, followed by creating 1 more sfc which will be pendingf.

e2e_test/e2e_test.go
```diff
@@ -461,6 +483,74 @@ var _ = g.Describe("E2E integration testing", g.Ordered, func() {
...
+				sfcNum := (dpuResNum / 2)
+
+				g.By("Should create N SFCs")
+				for i := 0; i < sfcNum; i++ {
+					sfc := sfcNew(i, imageRef)
+					testutils.SfcCreate(dpuSideClient, sfc)
+				}
+
+				g.By("Should get all N SFC Pods running")
+				for i := 0; i < sfcNum; i++ {
```
wizhaoredhat, line 506: I think the Marvell DPU only supports 1 SFC atm?

thom311: Yes. On the DPU side, the number `NoOfPortPairs` is hard coded to two (resulting in two `openshift.io/dpu` resources). And since a SFC Pod requires two such resources, only one SFC can be started at a time.

I think that should change, because for Marvell (currently) those resources are just veth pairs. There is no need to limit those resources (I think), or there is no need to keep them not much higher.

Anyway. The test doesn't care. It checks how many resources are available. It works fine with having only 2 resources (1 SFC).

PR #435, opened 2025-06-18
Bump CDA with incluster-registry deployment

SamD2021:


thom311:
/lgtm

Giving ACK, because the main review already happened when merging the CDA feature. I did not re-review that. Pulling in latest CDA is supposed to be a no-brainer (provided CI passes).

SamD2021:
There was a small bug that got in where it wouldn't wait enough for the default route for the in-cluster registry, so lets wait for it to pass with this hotfix and I'll get it merged on cda and bump cda to the proper version

PR #440, opened 2025-06-20
[th/marvell-no-extra-args] taskfiles: drop CDA_MARVELL_TOOLS_EXTRA_ARGS workaround for deploying marvell

thom311:
With [1], we no longer need to set the environment variable. Drop this.

[1] https://github.com/bn222/cluster-deployment-automation/pull/355

### DIFF
--- a/taskfiles/clusters.yaml
+++ b/taskfiles/clusters.yaml
@@ -22,13 +22,10 @@ tasks:
     internal: true
     deps:
       - task: prepare-e2e-test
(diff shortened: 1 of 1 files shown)


SamD2021:
Please bump with the new fix that was merged into CDA for the image registry, it should fix the errors you are getting in the e2e tests

bn222:
/close

https://github.com/openshift/dpu-operator/pull/461

PR #441, opened 2025-06-23
CI for redfish simplification, take two

swtaylor:


SamD2021:
I'm assuming this is just for testing the changes that you want merged into CDA so i'm adding a hold tag for us to not accidentally merge it here even though it passes everything

swtaylor:
@SamD2021 Nothing in any of my updates should've had any effect on marvell. Is that e2e just flaky? can we ignore those failures?

PR #445, opened 2025-06-26
fix(ci): validate venv functionality before reusing cached environments

SamD2021:
Motivation:
The CI would sometimes fall into state where it tries to use a package that should be installed in the venv but doesn't because the venv has already been invalidated.
The reason for this is that our caching check doesn't actually validate the venv and instead checks if we are still in the same revision for the submodule.
To fix this we should validate the venv properly while mainting proper caching when the venv is still valid.

Changes:
- Add comprehensive venv validation to check python/pip binaries and packages
- Prevent CI failures from broken cached venvs that pass SHA checks
- Use trap for robust temporary file cleanup in validation logic
- Maintain efficient caching while ensuring venv reliability

### DIFF
--- a/hack/prepare-venv.sh
+++ b/hack/prepare-venv.sh
@@ -2,34 +2,101 @@
 
 set -e
 
+validate_venv() {
+    local venv_path=$1
+    local requirements_file=$2
+    
+    # Check if essential binaries exist and are executable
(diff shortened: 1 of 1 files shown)


bn222:
/lgtm
/approve

Thanks @SamD2021 It's good that this is finally getting sorted.

PR #452, opened 2025-06-30
Improve log message on why test failed

bn222:
We're hitting a flake where the pod is not coming up on time during SFC creation. This PR improves the logging around the whole creation process by splitting it up in 2 Eventually blocks. First, the pod needs to be created. Next, it should eventually be in running state. If it's not, we need to know what state it is in do help debug this.


bn222:
known flake that we're working on fixing

/test make-e2e-test

PR #454, opened 2025-06-30
Draft Inital PTL code

wizhaoredhat:


internal/daemon/hostsidemanager.go
```diff
@@ -94,6 +94,14 @@ func NewHostSideManager(vsp plugin.VendorPlugin, opts ...func(*HostSideManager))
 		opt(h)
 	}
 
+	addr, port, err := h.vsp.Start()
+	if err != nil {
+		h.log.Error(err, "VSP Start() returned error")
+		return nil
+	}
+	h.addr = addr
+	h.port = port
```
bn222, line 103: moved here, why?

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -0,0 +1,367 @@
...
+		// This error may be fine. Maybe our host doesn't even run
+		// NetworkManager. Ignore.
+		klog.Infof("nmcli device set %s managed no failed with error %v", interfaceName, err1)
+	}
+
+	optimistic_dad_file := "/proc/sys/net/ipv6/conf/" + interfaceName + "/optimistic_dad"
+	err1 = os.WriteFile(optimistic_dad_file, []byte("1"), os.ModeAppend)
+	if err1 != nil {
+		klog.Errorf("Error setting %s: %v", optimistic_dad_file, err1)
+	}
```
bn222, line 136: This could be a function with a name like disableDuplicateAddressDetection

bn222: How was this done without the DPU operator? Do we ened this only for the vsp link?

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -0,0 +1,367 @@
...
+	version        string
+	isDPUMode      bool
+	dpuPcieAddress string
+}
+
+func SetSriovNumVfs(pciAddr string, numVfs int) error {
+	klog.Infof("SetSriovNumVfs(): set NumVfs device %s numvfs %d", pciAddr, numVfs)
+	numVfsFilePath := filepath.Join("/sys/bus/pci/devices", pciAddr, "sriov_numvfs")
+	bs := []byte(strconv.Itoa(numVfs))
+	err := os.WriteFile(numVfsFilePath, []byte("0"), os.ModeAppend)
```
bn222, line 61: Use `afero` instead to allow for testing.

internal/daemon/plugin/vendorplugin.go
```diff
@@ -180,13 +186,20 @@ func (gp *GrpcPlugin) deployVsp() error {
 	return nil
 }
 
-func NewGrpcPlugin(dpuMode bool, client client.Client, opts ...func(*GrpcPlugin)) (*GrpcPlugin, error) {
+func NewGrpcPlugin(dpuMode bool, dpuPciDevice *ghw.PCIDevice, client client.Client, opts ...func(*GrpcPlugin)) (*GrpcPlugin, error) {
+	var dpuPcieAddress string
```
bn222, line 190: We need a different identifier than pcie address at the vsp/grpc level. The assumption that this code makes is that we can't configure more devices. Some DPUs can be configured in a way that creates more PFs. There is a hierarchy at play here. At the vsp level, we need a way to identify groups of PCIe devices that belong to the same physical card. If we expose PCIe devices at this level in the API, then that detail (allowing to create multiple devices) leaks through the API. Can we use the serial number instead? This is the most important comment in my review.

internal/daemon/hostsidemanager.go
bn222: Also, error handling is not correct anymore.

wizhaoredhat: Because SetNumVf API is called before start if not moved here.

internal/daemon/plugin/vendorplugin.go
wizhaoredhat: I am not sure how we can use the serial number. The best I have found is using the BUS ID of the PCIe device.

I don't quite understand why using the PCIe details would cause harm. Maybe we need to take it off on a call.

internal/daemon/device-handler/dpu-device-handler/dpudevicehandler.go
```diff
@@ -101,15 +101,20 @@ func (d *dpuDeviceHandler) SetupDevices() error {
...
+		if d.dpuMode {
+			d.log.Info("Failed to set sriov numVFs, but ignoring error in DPU mode", "error", err)
+		} else {
+			return fmt.Errorf("failed to set sriov numVfs: %v", err)
+		}
+	} else {
+		if numVfs.VfCnt == 0 {
+			return fmt.Errorf("SetNumVfs ran, but numVfs == %d", numVfs.VfCnt)
+		}
+		d.log.Info("Num VFs set by VSP", "vf_count", numVfs.VfCnt)
```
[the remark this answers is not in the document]
wizhaoredhat: This is error handling that is all. We can technically set NumVfs on the PTL card. I actually want to try to make use of it this way.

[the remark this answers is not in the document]
bn222: ok

internal/daemon/plugin/vendorplugin.go
bn222: The PFs on the same DPU are linked, so we need to represent them as a unit. You can still perform configuration at the PF level, but if the DPU goes down, both PFs go down so they need to be treated as a unit or you will not know which pods to drain. I know that we are currently draining the whole host, but since this is an API level change, we should think about how to handle it when we have DRA too. In that case, we'll be able to drain all the pods that are "affected by a specific DPU going down". That's more than "affected by a specific PF going down".

Then, some DPUs can create (depending on the DPU level configuration) more PFs. For that reason, the devices are configured at the level of a DPU, not at the level of the PF. The DPU CRs should reflect the highest level we can configure at. We will still be able to configure at a lower level too, but we need the capability to configure at a higher level.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
wizhaoredhat: I am reusing logic from Marvell which I will clean up and make it a library. It is used for the VSP link.

internal/daemon/plugin/vendorplugin.go
wizhaoredhat: Done changed the API as discussed.

bn222:
Generally I agree with the intent and approach here. Thanks for correcting it.

To get this passing in the CI, we will need to perform an API change first. Let's create a PR called "Introduce DPU Identifier" which introduces the API change, and we align all the existing implementations so that our CI still passes. Then, we can enable PTL.

API changes are the hardest to do since it means that you need to get everyone to align on it, but since we need it, it needs to be done first.

wizhaoredhat:
This commit has been repurposed to handle Start() function sequencing.

wizhaoredhat:
```
openshift-dpu-operator                             dpu-daemon-k8m86                                            1/1     Running                  0               12s     192.168.122.218   worker-229            <none>           <none>
openshift-dpu-operator                             dpu-daemon-m44bw                                            1/1     Running                  0               12s     192.168.122.90    worker-229-ptl        <none>           <none>
openshift-dpu-operator                             dpu-operator-controller-manager-694f86768b-984g5            1/1     Running                  0               117s    10.128.3.148      worker-229            <none>           <none>
openshift-dpu-operator                             network-resources-injector-699988f484-wjjnz                 1/1     Running                  0               11s     10.128.3.149      worker-229            <none>           <none>
openshift-dpu-operator                             vsp-2sl8l                                                   1/1     Running                  0               8s      192.168.122.90    worker-229-ptl        <none>           <none>
openshift-dpu-operator                             vsp-fjgp4                                                   1/1     Running                  0               8s      192.168.122.218   worker-229            <none>           <none>
```
It works now

wizhaoredhat:
@bn222  please review

internal/daemon/dpusidemanager.go
```diff
@@ -57,7 +59,7 @@ func (s *DpuSideManager) DeleteBridgePort(context context.Context, bpr *pb.Delet
 	return &emptypb.Empty{}, err
 }
 
-func NewDpuSideManger(vsp plugin.VendorPlugin, config *rest.Config, opts ...func(*DpuSideManager)) *DpuSideManager {
+func NewDpuSideManager(vsp plugin.VendorPlugin, config *rest.Config, opts ...func(*DpuSideManager)) (*DpuSideManager, error) {
```
bn222, line 62: It's a bit weird that the `NewDpuSideManager` function also starts something, a bit unexpected even. That might cause some issues due to started vsp too early (at the time of calling `NewDpuSideManager` function). Can we have a separate start function so that the creation of the struct isn't tied together with the starting of a d.vsp.Start()

internal/daemon/hostsidemanager.go
```diff
@@ -94,11 +94,18 @@ func NewHostSideManager(vsp plugin.VendorPlugin, opts ...func(*HostSideManager))
 		opt(h)
 	}
 
+	addr, port, err := h.vsp.Start()
+	if err != nil {
+		return nil, fmt.Errorf("failed calling VSP Start(): %v", err)
+	}
+	h.addr = addr
+	h.port = port
+
```
bn222, line 103: Same remark here

internal/daemon/dpusidemanager.go
wizhaoredhat: Ok, I moved it into to Serve routine with stricter order. Let me know if this is better.

internal/daemon/hostsidemanager.go
wizhaoredhat: Done, with the other change.

wizhaoredhat:
Marvell tests pass before, but we are hitting some flakiness.

PR #467, opened 2025-07-04
Add Intel NetSec Accelerator Support with DPU Operator

wizhaoredhat:
This commit adds the intial PTL support for the DPU Operator. The goal of this commit is to have the daemon properly detect the Intel NetSec Accelerator. Then deploy the VSP properly. This commit deploys the operator in a 1 cluster mode. Which means all daemons and vsps run in 1 cluster for the host and the Intel NetSec Accelerator. There was changes to the taskfiles to achieve this.
    
Also remove duplication of code between Marvell and Intel NetSec by introducing the new vspnetutils and using platform and afero where appropriate.
    
For PTL we must use the PCIe Serial Number for the DPU Identifier. This is because the PTL has 2 PFs.

### DIFF
--- a/internal/daemon/plugin/vendorplugin.go
+++ b/internal/daemon/plugin/vendorplugin.go
@@ -25,14 +25,18 @@ import (
 //go:embed bindata/*
 var binData embed.FS
 
+type DpuIdentifier string
+
 const VspImageIntel string = "IntelVspImage"
 const VspImageMarvell string = "MarvellVspImage"
+const VspImageIntelNetSec string = "IntelNetSecVspImage"
 
 const VspImageP4Intel string = "IntelVspP4Image"
 
 var VspImages = []string{
 	VspImageIntel,
 	VspImageMarvell,
+	VspImageIntelNetSec,
 	// TODO: Add future supported vendor plugins here
 }
 
@@ -78,16 +82,17 @@ type VendorPlugin interface {
 }
 
 type GrpcPlugin struct {
-	log         logr.Logger
-	client      pb.LifeCycleServiceClient
-	k8sClient   client.Client
-	opiClient   opi.BridgePortServiceClient
-	nfclient    pb.NetworkFunctionServiceClient
-	dsClient    pb.DeviceServiceClient
-	dpuMode     bool
-	vsp         VspTemplateVars
-	conn        *grpc.ClientConn
-	pathManager utils.PathManager
+	log           logr.Logger
+	client        pb.LifeCycleServiceClient
+	k8sClient     client.Client
+	opiClient     opi.BridgePortServiceClient
+	nfclient      pb.NetworkFunctionServiceClient
+	dsClient      pb.DeviceServiceClient
+	dpuMode       bool
+	dpuIdentifier DpuIdentifier
+	vsp           VspTemplateVars
+	conn          *grpc.ClientConn
+	pathManager   utils.PathManager
 }
 
 func NewVspTemplateVars() VspTemplateVars {
@@ -127,22 +132,23 @@ func (g *GrpcPlugin) Start() (string, int32, error) {
 		err := g.ensureConnected()
 		if err != nil {
 			if time.Since(start) >= timeout {
-				return "", 0, fmt.Errorf("Failed to ensure GRPC connection after %v: %v", timeout, err)
+				return "", 0, fmt.Errorf("failed to ensure GRPC connection after %v: %v", timeout, err)
 			}
 			time.Sleep(interval)
 			continue
 		}
 
-		ipPort, err := g.client.Init(context.TODO(), &pb.InitRequest{DpuMode: g.dpuMode})
+		ipPort, err := g.client.Init(context.TODO(), &pb.InitRequest{DpuMode: g.dpuMode, DpuIdentifier: string(g.dpuIdentifier)})
 		if err != nil {
 			if time.Since(start) >= timeout {
-				return "", 0, fmt.Errorf("Failed to start serving after %v: %v", timeout, err)
+				return "", 0, fmt.Errorf("failed to start serving after %v: %v", timeout, err)
 			}
 			time.Sleep(interval)
 			continue
 		}
 
-		g.log.Info("Init succeeded", "duration", time.Since(start))
+		g.log.Info("GrpcPlugin Start() succeeded", "duration", time.Since(start), "ip", ipPort.Ip, "port", ipPort.Port, "dpuMode",
+			g.dpuMode, "dpuIdentifier", g.dpuIdentifier)
 		return ipPort.Ip, ipPort.Port, nil
 	}
 }
@@ -196,13 +202,14 @@ func (gp *GrpcPlugin) deployVsp() error {
 	return nil
 }
 
-func NewGrpcPlugin(dpuMode bool, client client.Client, opts ...func(*GrpcPlugin)) (*GrpcPlugin, error) {
+func NewGrpcPlugin(dpuMode bool, dpuIdentifier DpuIdentifier, client client.Client, opts ...func(*GrpcPlugin)) (*GrpcPlugin, error) {
 	gp := &GrpcPlugin{
-		dpuMode:     dpuMode,
-		vsp:         VspTemplateVars{},
-		k8sClient:   client,
-		log:         ctrl.Log.WithName("GrpcPlugin"),
-		pathManager: *utils.NewPathManager("/"),
+		dpuMode:       dpuMode,
+		dpuIdentifier: dpuIdentifier,
+		vsp:           VspTemplateVars{},
+		k8sClient:     client,
+		log:           ctrl.Log.WithName("GrpcPlugin"),
+		pathManager:   *utils.NewPathManager("/"),
 	}
 
 	for _, opt := range opts {
--- a/internal/daemon/vendor-specific-plugins/intel-netsec/main.go
+++ b/internal/daemon/vendor-specific-plugins/intel-netsec/main.go
@@ -1,0 +1,341 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"sync"
+
+	"github.com/go-logr/logr"
+	"github.com/jaypipes/ghw"
+	pb "github.com/openshift/dpu-operator/dpu-api/gen"
+	"github.com/openshift/dpu-operator/internal/daemon/plugin"
+	vspnetutils "github.com/openshift/dpu-operator/internal/daemon/vendor-specific-plugins/common"
+	"github.com/openshift/dpu-operator/internal/platform"
+	"github.com/openshift/dpu-operator/internal/utils"
+	opi "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
+	"github.com/spf13/afero"
+	"go.uber.org/zap/zapcore"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const (
+	Version                              string = "0.0.1"
+	IPv6AddrDpu                          string = "fe80::1"
+	IPv6AddrHost                         string = "fe80::2"
+	DefaultPort                          int32  = 8085
+	IntelVendorID                        string = "8086"
+	IntelNetSecHostVfDeviceID            string = "1889" // Intel Corporation Ethernet Adaptive Virtual Function
+	IntelNetSecDpuSFPf0PCIeAddress       string = "0000:f4:00.0"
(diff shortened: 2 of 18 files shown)


internal/daemon/vendor-specific-plugins/marvell/main.go
```diff
@@ -672,7 +673,7 @@ func (vsp *mrvlVspServer) configureIP(dpuMode bool) (pb.IpPort, error) {
 		return pb.IpPort{}, err
 	}
 	klog.Infof("Interface Name: %s", ifName)
-	err = enableIPV6LinkLocal(ifName, addr)
+	err = vspnetutils.EnableIPV6LinkLocal(vsp.fs, ifName, addr)
```
bn222, line 676: I like the abstraction but I'm wondering if the vspnetutils is the right abstraction here. It seems that it contains multiple smaller useful things, almost as if we might want to abstract different partitions of vspnetutils. Don't necessarily change it, but some discussion is needed as this is fundamental and at and architectural level. Again, might be that we call it a day today, and add it to the backlog to split it up if we see the benefit. It's about determining the right names to express the intent.

internal/platform/marvell-dpu.go
```diff
@@ -28,9 +28,7 @@ func (d *MarvellDetector) Name() string {
 	return d.name
 }
 
-// IsDPU checks if the PCI device Attached to the host is a Marvell DPU
-// It returns true if device has Marvell DPU
```
bn222, line 32: still checks if it's a marvell dpu. I presume you removed the incorrect comment after realizing that this returns true when it runs ON the dpu?

internal/platform/marvell-dpu.go
```diff
@@ -56,11 +54,16 @@ func (pi *MarvellDetector) IsDpuPlatform(platform Platform) (bool, error) {
...
+	// TODO: Implement a way to get the DPU identifier.
+	return "", nil
+}
+
+func (pi *MarvellDetector) VspPlugin(dpuMode bool, vspImages map[string]string, client client.Client, pm utils.PathManager, dpuIdentifier plugin.DpuIdentifier) (*plugin.GrpcPlugin, error) {
 	template_vars := plugin.NewVspTemplateVars()
 	template_vars.VendorSpecificPluginImage = vspImages[plugin.VspImageMarvell]
 	template_vars.Command = `[ "/vsp-mrvl" ]`
-	return plugin.NewGrpcPlugin(dpuMode, client, plugin.WithVsp(template_vars), plugin.WithPathManager(pm))
+	return plugin.NewGrpcPlugin(dpuMode, dpuIdentifier, client, plugin.WithVsp(template_vars), plugin.WithPathManager(pm))
```
bn222, line 66: instead of a round trip first asking the dpu identifier with `GetDpuIdentifier` and then passing it into VspPlugin where it's the only one used, is there a reason to expose this to the caller of `GetDpuIdentifier`?

internal/platform/platform.go
```diff
@@ -24,13 +31,55 @@ func (hp *HardwarePlatform) PciDevices() ([]*ghw.PCIDevice, error) {
...
+		return nil, err
+	}
+	return netInfo.NICs, nil
+}
+
 func (hp *HardwarePlatform) Product() (*ghw.ProductInfo, error) {
 	return ghw.Product()
 }
 
+func (hp *HardwarePlatform) ReadDeviceSerialNumber(pciDevice *ghw.PCIDevice) (string, error) {
```
bn222, line 46: nicely done. Any reason to pass in a pointer? Yes it costs a copy in a garbage collected language, but the price is that I don't know for sure if pciDevice is only an input or also an output.

internal/platform/platform.go
```diff
@@ -24,13 +31,55 @@ func (hp *HardwarePlatform) PciDevices() ([]*ghw.PCIDevice, error) {
...
+	file, err := os.Open(devicePath)
+	if err != nil {
+		return "", fmt.Errorf("failed to open config space: %v", err)
+	}
+	defer file.Close()
+
+	// Seek to offset 0x150
+	// Capabilities: [150] Device Serial Number 88-dc-97-ff-ff-44-24-8b
+	const serialOffset = 0x150
+	_, err = file.Seek(serialOffset, 0)
```
bn222, line 62: ghw.PCIDevice didn't have this capability?

internal/platform/vendordetector.go
```diff
@@ -139,7 +98,7 @@ func (d *DpuDetectorManager) Detect(vspImages map[string]string, client client.C
 		}
 
 		if dpuPlatform {
-			vsp, err := detector.VspPlugin(true, vspImages, client, pm)
+			vsp, err := detector.VspPlugin(true, vspImages, client, pm, "")
```
bn222, line 101: Is it the case that the last arg is always "" if first is true?

wizhaoredhat: Yes, because we don't have a DPU Identifier for running on the DPU. It shouldn't be needed since there will only be 1 VSP and 1 daemon running on the DPU. There shouldn't be anymore.

internal/platform/platform.go
wizhaoredhat: Unfortunately no... It doesn't read extended capabilities.

internal/platform/platform.go
wizhaoredhat: Personal preference. In C++ it would be static const pointer but we don't have that in golang.... 

I like pointers rather than copies. Would a function commentary suffice?

internal/daemon/vendor-specific-plugins/marvell/main.go
wizhaoredhat: Hmmm, When it grows larger, we can move it. If the daemon needs these functions that we move it to a more common place. But I think VSP is appropriate right now with the current architecture.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	for _, opt := range opts {
+		opt(vsp)
+	}
+
+	return vsp
+}
+
+func main() {
```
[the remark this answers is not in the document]
wizhaoredhat: I thought about this. I think we want to have a VSP running and using the existing architecture because the more branches we have in the code, the more complicated our testing would need to be. 
E.g. If VSP works with Marvell and IPU, it will work for PTL.
I was tempted to move all of this into the daemon since it is generic and we will be using only up-streamed components for PTL.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -0,0 +1,337 @@
...
+
+	ipPort, err := vsp.configureIP(in.DpuMode)
+
+	return &pb.IpPort{
+		Ip:   ipPort.Ip,
+		Port: ipPort.Port,
+	}, err
+}
+
+// TODO: Implement this
```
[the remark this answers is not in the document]
wizhaoredhat: It's not 100% complete yet. For PTL, we may need to add VETH devices for GetDevices. It's still WIP, the whole devices and SFC stuff.

internal/daemon/vendor-specific-plugins/marvell/main.go
bn222: Ok

internal/platform/marvell-dpu.go
bn222: Ok

internal/platform/vendordetector.go
bn222: Right. Sounds ok

taskfile.yaml
```diff
@@ -115,6 +123,24 @@ tasks:
       - KUBECONFIG="/root/kubeconfig.microshift" oc -n openshift-dpu-operator wait --for=condition=ready pod --all --timeout=300s
       - KUBECONFIG="/root/kubeconfig.ocpcluster" oc -n openshift-dpu-operator wait --for=condition=ready pod --all --timeout=300s
 
+  deploy-1c:
```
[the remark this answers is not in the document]
bn222: Ok even better then. That means we also support ipu with 1c. Great work.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -0,0 +1,337 @@
...
+		Level:       zapcore.DebugLevel,
+	}
+	options.BindFlags(flag.CommandLine)
+	flag.Parse()
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&options)))
+	vsp := &intelNetSecVspServer{
+		log:         ctrl.Log.WithName("IntelNetSecVsp"),
+		pathManager: *utils.NewPathManager("/"),
+		done:        make(chan error),
+		fs:          afero.NewOsFs(),
```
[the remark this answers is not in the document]
bn222: Ok if you add a backlog jira item, then we can close it

internal/daemon/plugin/vendorplugin.go
```diff
@@ -78,16 +82,17 @@ type VendorPlugin interface {
...
-	conn        *grpc.ClientConn
-	pathManager utils.PathManager
+	log           logr.Logger
+	client        pb.LifeCycleServiceClient
+	k8sClient     client.Client
+	opiClient     opi.BridgePortServiceClient
+	nfclient      pb.NetworkFunctionServiceClient
+	dsClient      pb.DeviceServiceClient
+	dpuMode       bool
+	DpuIdentifier DpuIdentifier
```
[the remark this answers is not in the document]
wizhaoredhat: I can make this change.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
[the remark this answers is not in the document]
wizhaoredhat: Will create backlog Jira.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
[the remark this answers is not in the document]
wizhaoredhat: To keep the same architecture and conventions. Where there is a VSP pod running. I think we should keep it as is. For example when writing E2E scripts that look at the components agnostically (expecting daemon + vsp). Having a separate VSP container/binary follows the convention. I rather keep the convention for now, and optimize in the future.

If IPU ovs-p4 gets upstreamed and Marvell components such as cp_agent have an RPM. Then we could theoretically merged everything in to the daemon code.

internal/platform/platform.go
wizhaoredhat: Should be ok... Maybe golang will support const pointers which would make alot of sense from a programming language point of view.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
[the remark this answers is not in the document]
wizhaoredhat: https://issues.redhat.com/browse/IIC-729

internal/platform/marvell-dpu.go
wizhaoredhat: It needs to do this round trip because DPU Identifier is used to detect other PFs on the same DPU
```
		var dpuDevices []plugin.DpuIdentifier
		for _, pci := range devices {
			isDpu, err := detector.IsDPU(d.platform, *pci, dpuDevices)
			if err != nil {
				return false, nil, errors.Errorf("Error detecting if device is DPU with detector %v: %v", detector.Name(), err)
			}
			if isDpu {
				identifier, err := detector.GetDpuIdentifier(d.platform, pci)
				if err != nil {
					return false, nil, errors.Errorf("Error getting DPU identifier with detector %v: %v", detector.Name(), err)
				}
				dpuDevices = append(dpuDevices, identifier)
				vsp, err := detector.VspPlugin(false, vspImages, client, pm, identifier)
				if err != nil {
					return true, nil, err
				}
				return false, vsp, nil
			}
		}
```

wizhaoredhat: Just having in the VspPlugin() method is not enough. We need to pass it in.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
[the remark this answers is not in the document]
bn222: What if only ipu ovs-p4 gets upstreamed? Then, we would not run a vsp in a pod in case of ipu but still run a vsp when we have the marvell. To run a vsp in a pod or as a goroutine in the daemon is a matter of which dpu you're running on.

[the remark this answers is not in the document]
wizhaoredhat: We might want to keep the VSP as a separate pod architecture in. If we want to support the complete lifecycle there will be some proprietary tools needed. For example to upgrade the firmware on the E823 interfaces requires a special binary from Intel. I suspect most vendors would have such a tooling for firmware.

Supporting firmware could be made part of the operator and part of the lifecycle of the device.

I think the architecture is good and we shouldn't diverge from it.

wizhaoredhat:
Marvell test before. But we have an issue with the server flakiness which is being solved in CDA.

PR #471, opened 2025-07-08
[th/run-e2e-test-suite] e2e-test: add "run-e2e-test-suite.sh" script

thom311:
Benefits:

- when the primary/first run of the test fails, the script will re-run the test a few times more. This is useful to see whether the test would pass, if it would run again (e.g. a fluke). In any case, the overall run will be a failure, even if all the re-runs pass.

- environment variables like EXTERNAL_CLIENT_DEV can be set by the caller to overwrite the settings (e.g. to test on a cluster with a different setup).

- supports TEST_FOCUS variable.

- hack/run-e2e-test-suite.sh can be run directly, without the task file. In particular, you can run only the e2e-test suite, without the traffic-flow-tests (unlike `task run-e2e-test`).

### DIFF
--- a/hack/run-e2e-test-suite.sh
+++ b/hack/run-e2e-test-suite.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+set -e
+
+die() {
+    printf '%s\n' "$*"
+    exit 1
+}
+
+cd "$(dirname "$0")/.." || die "failure to change directory for $0"
+
+export FAST_TEST="${FAST_TEST:-true}"
+export REGISTRY="${REGISTRY:-$(hostname | sed 's/$/:5000/')}"
+export NF_INGRESS_IP="${NF_INGRESS_IP:-10.20.30.2}"
+export EXTERNAL_CLIENT_DEV="${EXTERNAL_CLIENT_DEV:-eno12409}"
+export EXTERNAL_CLIENT_IP="${EXTERNAL_CLIENT_IP:-10.20.30.100}"
+export BINDIR="${BINDIR:-bin}"
+export ENVTEST_K8S_VERSION="${ENVTEST_K8S_VERSION:-$(sed -n 's/^ *ENVTEST_K8S_VERSION: \+\(.*\)$/\1/p' taskfile.yaml)}"
+
+BINDIR_ABS="$BINDIR"
+if [[ "$BINDIR_ABS" != /* ]] ; then
+    BINDIR_ABS="$PWD/$BINDIR"
+fi
+
+GINKO_ARGS=()
+
+if [ -n "$TEST_FOCUS" ] ; then
+    GINKO_ARGS+=( "-focus=${TEST_FOCUS}" )
+fi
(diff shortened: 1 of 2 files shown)


hack/run-e2e-test-suite.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo ">>> Test failed. Rerun a few times!!"
+    for i in {1..5} ; do
+        echo ">>> Re-Run #$i..."
+        rc2=0
+        run_test || rc2=0
+        echo ">>> Re-Run #$i completed: result $rc2"
+    done
+fi
+
+exit "$rc"
```
bn222, line 51: ginkgo doesn't have something for this?

bn222: While retries can sometimes obscure the true problem, our goal is to diagnose and resolve the root cause of failures, not just mask them. Instead of simply retrying, we need to improve our logging. This will give us a much clearer picture when an error occurs.

If we merge this patch, any subsequent patch that makes the CI more flaky will be allowed to go in. We spent a lot of time to make it less flaky and this patch opens the floodgates in the opposite direction.

bn222: I now see that the loop is a bit non-standard. Instead of the typical "retry a few times or return true if hit a good run", it always returns the first run but still tries to run it a few times. Another reviewer (see slack) also had concerns about this wrt runtime. 

What do we gain really? We have our CI capturing bad runs. Where is the logging lagging to figure out what's wrong?

1. Inconsistent Success Criteria: If the first run fails, why retry at all if you're going to fail anyway?
2. Confusing Signals: Developers will see multiple test outputs but won't know which result actually determined the pass/fail status 
3.Language Inconsistency: The project already has robust Go-based testing infrastructure with Ginkgo/Gomega, making the shell script an outlier
4. Maintenance Burden: Shell scripts are harder to maintain, debug, and integrate with the existing test ecosystem

If the intent is to gather more information about a failure (which is a way to inch towards devising a fix), add code in go to capture more information so that any test can reuse a helper that does that.

thom311: > IIUC the only info you want to know is that if we retry running this after it failed, that it then succeeds?

From the re-run loop? Yes.

Beyond that, the patch adds a script with several benefits listed in the commit message.

thom311: I mean, run the entire test. E.g. the "setup" portion of `g.BeforeAll()`. All the code that leads up to the `Eventually()`. Just repeating the Eventually() is not useful in general(*)

(*) unless case where you have a tight timeout, then when the timeout is missed, you run the `Eventuall()` bit longer to see whether it would complete if the timeout were longer.

Maybe that should be done for IIC-651. If the SFC doesn't start the pod in 1 minute, check whether it would have started in a few more minutes, and whether the problem is that sometimes pod running takes an unexpected log time. That would be useful to see. But is something different from what is aimed here.

PR #302, opened 2025-02-13
bump to go 1.23.4

SalDaniele:

PR #474, opened 2025-07-09
[OCPBUGS-57962] [th/bump-4.20]

thom311:
https://issues.redhat.com/browse/OCPBUGS-57962

see also the bump to 4.19: 443a69cc5f8f546f986c447a5eeffb21ff48b9ca


thom311:
> /unhold
> 
> You missed all the go.mod files:
> 
> ```
> -go 1.23.6
> +go 1.24.4
> ```

does that need to be part of this PR? You can compile a project that asks for 1.23.6 with a 1.24.4 compiler (can you not?). This PR is large already, and introducing additional changes delays it.

thom311:
> You can compile a project that has go 1.23.6 with the 1.24.4 compiler but you will get the behavior of the new compiler. 

golang has strong commitments towards backward compatibility. The go version in `go.mod` specifies the minimum version required (and specifies the language features that are expected). Bumping this is unnecessary and actually wrong. We bump, when we need new language features, not because we now use a newer compiler.

Of course, we would expect that bumping the version would not introduce problems. If 1.24 compiler were unable to compile the project, we would notice either way. In practice, 1.24 will compile this source just fine, regardless what is specified in `go.mod`.

Bumping it is still cumbersome to do right now. For example, we would need to sync with Marvell about the base images that they use. I think they do not use the ones from `registry.ci.openshift.org`, but patch the Dockerfiles. Pushing this now, might cause inconveniences and break things.

For example, we bumped to 1.23 in https://github.com/openshift/dpu-operator/pull/302. The history does not reveal why that bump was done (such things should be recorded), but it seems unrated to the base images in the Dockerfile. I suspect, it was related to vendoring in some dependencies that required 1.23.

---

> In addition, we would need to bump OCP version in CDA too but that can be another PR (or it can be in this PR).

Let's do that later. Switching to an different, unstable OCP version is probably encountering several issues and requires separate effort.

thom311:
anyway. Added a commit to bump the go version to move this forward.

PR #482, opened 2025-07-18
Add support for Veth Devices and SR-IOV devices with VLANs

wizhaoredhat:
This commit prepares the PTL VSP plugin towards supporting and OvS dataplane.


cluster-deployment-automation
thom311, line 1: this change to CDA git-subtree is probably wrong.

internal/daemon/device-handler/dpu-device-handler/dpudevicehandler.go
```diff
@@ -86,12 +86,6 @@ func (d *dpuDeviceHandler) SetupDevices() error {
 
 	defer close(d.setupDevicesDone)
 
-	// Currently NF devices do not require any setup outside the VSP
-	if d.dpuMode {
-		d.log.Info(("Dpu mode detected, skipping devHandler devices setup"))
-		return nil
-	}
```
thom311, line 93: It doesn't seem right to proceed to setup VFs (at least not for some kinds of DPU).

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
...
+	//nfInterfaceName is the name of the interface on the Network Function attached to the container
+	//dpInterfaceName is the name of the interface on the Data Plane OvS side
+	nfInterfaceName := fmt.Sprintf("nf_if%d", index)
+	dpInterfaceName := fmt.Sprintf("dp_if%d", index)
+
+	var nfLink netlink.Link
+	var dpLink netlink.Link
+	var err error
+
+	// Create early for clean up purposes.
```
thom311, line 147: This comment seems unnecessary and explains nothing that is not already visible in the code. `deviceInfo` is used in the very next line with `DestroyVethPair(&deviceInfo)`.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
...
+		DpMAC:           "",
+	}
+
+	// Destroy existing veth pair if it exists
+	// This is to ensure that we do not have any stale veth pairs lying around.
+	// Error is fine if there is no existing veth pair.
+	_ = DestroyVethPair(&deviceInfo)
+
+	nfLink, _ = netlink.LinkByName(nfInterfaceName)
+	if nfLink != nil {
```
thom311, line 160: we won't find the interface here, if we call `DestroyVethPair` above.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
thom311: `DestroyVethPair()` only uses `dev.NfInterfaceName` (which probably makes sense). As it is, `DestroyVethPair()` would be better to only take a `peer_ifname string` argument. Then you can call `DestroyVethPair(nfInterfaceName)`.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
...
+		nfLink, err = netlink.LinkByName(nfInterfaceName)
+		if err != nil {
+			klog.Errorf("Error occurred in getting Link By Name of NF Link: %v", err)
+			return nil, err
+		}
+		dpLink, err = netlink.LinkByName(dpInterfaceName)
+		if err != nil {
+			klog.Errorf("Error occurred in getting Link By Name of DP Link: %v", err)
+			return nil, err
+		}
```
thom311, line 183: marvell code also creates Veth Pairs. It should re-use this helper -- which presumably does a better job at error handling and special cases..

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
 
 	return "", fmt.Errorf("network device not found for PCI address %s", pcieAddress)
 }
+
+// CreateVethPair function to create a veth pair with the given index and InterfaceInfo
+func CreateVethPair(index int) (*VEthPairDeviceInfo, error) {
```
thom311, line 137: this is a general purpose utility (that mainly only depends on netlink). It is good that it is under "common". But it should not be under "vendor-specific-plugins". There should be a utility package that can be used from anywhere (and in this case, the package mainly only depends on netlink).

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
...
+	//nfInterfaceName is the name of the interface on the Network Function attached to the container
+	//dpInterfaceName is the name of the interface on the Data Plane OvS side
+	nfInterfaceName := fmt.Sprintf("nf_if%d", index)
+	dpInterfaceName := fmt.Sprintf("dp_if%d", index)
```
thom311, line 141: I think there should be a function `CreateVethPair()` that takes the interface name and the peer name.

Choosing a particular name (nf_if%d) should be done by the caller, or a wrapper function `CreateNfVethPair()`.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
...
+		if errOnDestroy != nil {
+			klog.Errorf("Error occurred in destroying existing veth pair: %v", err)
+			err = errors.Join(errOnDestroy, err)
+		}
+		return nil, err
+	}
+
+	deviceInfo.DpMAC = dpLink.Attrs().HardwareAddr.String()
+
+	return &deviceInfo, nil
```
thom311, line 208: as a general purpose function, `CreateVethPair()` should return two `netlink.Link` instances (or an error).

If some upper layer wants to put some parts of that into a `deviceInfo` tuple, they can do so themselves. What I mean is that (general purpose) `CreateVethPair` should not return a less useful `deviceInfo`, but rather the netlink.Link that it already has. And any caller can do with them what they want.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -110,3 +132,134 @@ func GetNetDevNameFromPCIeAddr(platform platform.Platform, pcieAddress string) (
...
+	}
+
+	return nil
+}
+
+// SetSriovVlanId sets the VLAN ID for a specific virtual function (VF)
+// As per ip-link (8) VLAN is special where it disable VLAN tagging and filtering
+// Also incoming traffic will be filtered for the specific VLAN ID and will
+// have all VLAN tags stripped before being passedto the VF.
+func SetSriovVlanId(pfName string, vfId int, vlanId int) error {
```
thom311, line 230: this function basically calls `LinkSetVfVlan()` (plus the convenience to operate on an ifname, instead of a `netlink.Link`).

But you choose a new name for this convenience wrapper (around `LinkByName()`+`LinkSetVfVlan()`). I think the name should strongly resemble the underlying `LinkSetVfVlan` (unless that name is particularly bad for some reason). Maybe `LinkSetVfVlanByName()`. The benefit is that if you are familiar with the netlink API, then you can guess what a similarly named function does. 

Also, this function operates on any netdev -- at least in principle. The `pfName` parameter should be called something more general, like `ifname`, or `iface`, or whatever we commonly choose to call the names of netdev devices). `netlink.LinkByName()` calls this parameter just `name` (which I think is not good).

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
thom311: I think the patch should be split in a part that introduces new helper API, and a part that does something about PTL code.

internal/daemon/device-handler/dpu-device-handler/dpudevicehandler.go
wizhaoredhat:
```
	numVfs, err := d.vsp.SetNumVfs(8)
	if err != nil {
		// Currently NF devices do not require any setup outside the VSP
		// ignore the error if we are in DPU mode.
		if d.dpuMode {
			d.log.Info("Failed to set sriov numVFs, but ignoring error in DPU mode", "error", err)
		} else {
			return fmt.Errorf("failed to set sriov numVfs: %v", err)
		}
	} else {
		if numVfs.VfCnt == 0 {
			return fmt.Errorf("SetNumVfs ran, but numVfs == %d", numVfs.VfCnt)
		}
		d.log.Info("Num VFs set by VSP", "vf_count", numVfs.VfCnt)
	}
```
Below explains this. The API should be generic for all DPUs. And the API must be called, it is up to the VSP to properly respond back.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
wizhaoredhat: I think it is a better API to pass the data structure that represents VEthPairs, such that the objects can be tracked and followed everywhere without exceptions.

wizhaoredhat: I can remove that comment

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
wizhaoredhat: We can move the package once we see that the functions we created under vender specific plugin is going to be used elsewhere. Better not assume anything about future use when we can easily move it later.

wizhaoredhat: I think the patch is small enough that we don't need to split it up for review.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
wizhaoredhat: Good point. Changing this.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
wizhaoredhat: netlink.Link can become stale  I believe. It is better that the callers derive it from a fresh netlink call.

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
wizhaoredhat: DestroyVethPair can be used outside which passing the object in would be better.

wizhaoredhat:
@thom311 is this PR good to be merged?

PR #484, opened 2025-07-22
IPU plugin version update

arunvelayutham:
This version addresses slow progression of Host and IPU pod creation/deletion flow due to extended delays in P4 rule updates at IPU.

We now derive the VSI and vPort directly from the VF’s MAC address. Previously, we queried the IMC in real time during P4 rule programming, which added latency to rule add/delete APIs and slowed NF and host pod provisioning.

This update eliminates those frequent IMC calls for VFs, while we continue to query IMC for PHY interfaces (primary/secondary networks) since their MACs can be statically configured via NVMe.

Refer IIC-641 for more context.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/go-task/task/v3 v3.43.3
 	github.com/golang/glog v1.2.4
 	github.com/gorilla/mux v1.8.1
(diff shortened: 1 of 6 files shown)


arunvelayutham:
It needs "approved" label as well.

PR #485, opened 2025-07-22
Add OvS Support for the PTL

wizhaoredhat:
- This commit makes sure that spoof check is turned off and trust is turned on for the SR-IOV interfaces
- Then it takes those SR-IOV interfaces and plumbs it into an OvS bridge
- In order to plumb the correct interface, the VSP needs to convert the OPI API bridge port name to PF/VF indexes.

Additionally log levels were reduced down to prevent flooding of logs.

Also workaround issue with e810/e823 interfaces are not VLAN aware. Intel NetSec Accelerator e810 interfaces are not VLAN-aware in VEB mode, thus packets can be forwarded to MAC addresses across VLAN boundaries. This violates the IEEE802.1Q forwarding rules.
    
This issue is tracked here: https://issues.redhat.com/browse/RHEL-104927
    
To workaround this we need to set the host side to vepa mode. This would bypass the loopback on the bridge and the bridging features entirely. Intel is working on a fix to make the ethernet controllers become proper VLAN aware bridges.

### DIFF
--- a/internal/daemon/vendor-specific-plugins/common/vspnetutils.go
+++ b/internal/daemon/vendor-specific-plugins/common/vspnetutils.go
@@ -28,12 +28,16 @@ type VEthPairDeviceInfo struct {
 	PeerIfMAC string // MAC address of the Peer veth interface
 }
 
-type VfDeviceInfo struct {
+type VfDeviceKey struct {
 	PfInterfaceName string // Name of the PF interface
 	Id              int    // VF ID starting from 0
-	PciAddress      string // PCI address of the VF
-	Vlan            int    // VLAN ID of the VF (0 is not vlan tagged)
-	Allocated       bool   // Indicates if the VF is allocated or not
+}
+
+type VfDeviceInfo struct {
+	VfKey      VfDeviceKey
+	PciAddress string // PCI address of the VF
+	Vlan       int    // VLAN ID of the VF (0 is not vlan tagged)
+	Allocated  bool   // Indicates if the VF is allocated or not
 }
 
 // SetSriovNumVfs sets the number of virtual functions (VFs) for a given PCI address.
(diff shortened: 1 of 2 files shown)


internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -248,15 +290,100 @@ func (vsp *intelNetSecVspServer) GetDevices(ctx context.Context, in *pb.Empty) (
...
+		vsp.log.Error(err, "Error getting netdev name from PCIe address", "PCIeAddress", backPlanePcieAddr)
+		return nil, err
+	}
+
+	// For Intel NetSec, PF ID is always 0 and VF ID is mapped one to one.
+	key := vspnetutils.VfDeviceKey{
+		PfInterfaceName: pfIfName,
+		Id:              vfId,
+	}
+	return vsp.vfDevs[key], nil
```
thom311, line 335: are we sure that this key exists inside `vsp.vfDevs`? Why? That is not clear to me, maybe add a code comment. Otherwise, handle the potentially missing key to return an error.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -302,20 +429,40 @@ func (vsp *intelNetSecVspServer) setVlanIds(numVfs int) error {
...
 
+		// Disable spoof check for the VF to allow forwarding traffic.
+		err = vspnetutils.LinkSetVfSpoofChkByName(ifName, vfID, false)
+		if err != nil {
+			vsp.log.Error(err, "Error disabling spoof check for VF", "VFID", vfID, "InterfaceName", ifName)
+			return err
+		}
+
+		// Enable trust for the VF to allow forwarding traffic.
+		err = vspnetutils.LinkSetVfTrustByName(ifName, vfID, true)
```
thom311, line 440: note that we now have these ByName() helpers. They all basically lookup the link instance, and then call a netlink function on them.

makes me wonder, whether those helpers are really useful? Performance-wise now, but that doesn't matter here. But with these helper it also ends up being more code, then just calling `netlink.LinkByName()` once and use it with `netlink.LinkSet*()`. Are they really beneficial?

You talked about stale link entries, but if you lookup the link entre at the beginning of the loop and then use it 3 times, then there is no problem.

wizhaoredhat: I decided to change it.

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
```diff
@@ -248,15 +290,100 @@ func (vsp *intelNetSecVspServer) GetDevices(ctx context.Context, in *pb.Empty) (
 	}, nil
 }
 
-// TODO: Implement this
+// getVFName function to get the VF Name of the given OPI BridgePortName on DPU
+func (vsp *intelNetSecVspServer) getConnectedVf(OPIBridgePortName string) (*vspnetutils.VfDeviceInfo, error) {
+	// Use regex to get VFId from BridgePortName ex: host0-7
+	re := regexp.MustCompile(`host(\d+)-(\d+)`)
```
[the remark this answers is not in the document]
wizhaoredhat: Done

internal/daemon/vendor-specific-plugins/intel-netsec/main.go
wizhaoredhat: Done

internal/daemon/vendor-specific-plugins/common/vspnetutils.go
```diff
@@ -258,3 +298,27 @@ func VfPCIAddressFromVfIndex(fs afero.Fs, pfName string, vfId int) (string, erro
 	}
 	return pciAddress, err
 }
+
+// TODO: Use https://github.com/ovn-kubernetes/libovsdb/tree/main
+func CreateBridge(bridgeName string) error {
```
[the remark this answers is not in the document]
wizhaoredhat: We should have a period of time when we can do refactors like this once there is a need. We shouldn't have a reluctance to accept refactoring patches. It should definitely be done.

Right now all the functions are to be used by the VSP exclusively since it handles bridges and VFs. The daemon code (at the moment) doesn't handle this.

I see your concerns about refactoring, but we need to refactor when appropriate for sure!

I have added OvS to the name.

wizhaoredhat:
@bn222 PTAL once you are back

PR #488, opened 2025-07-28
IIC-711 Replace VSP P4 dockerfile with the P4 dockerfile

saynb:


wizhaoredhat:
LGTM
@bn222 PTAL.
We can override marvell tests

PR #489, opened 2025-07-30
Unify image handling

bn222:


internal/images/manager.go
```diff
@@ -0,0 +1,20 @@
+package images
+
+import "fmt"
```
thom311, line 3: this file seems to provide little benefit. It should be merged with `images.go`

hack/traffic_flow_tests.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-bash hack/prepare-venv.sh
+# bash hack/prepare-venv.sh
```
thom311, line 5: this change seems wrong

PR #490, opened 2025-07-31
IIC-763 Ignore iputils when installing nmcli as it is a weak dep

saynb:


Dockerfile.IntelVSP.rhel
```diff
@@ -18,7 +18,7 @@ WORKDIR /
 # dependencies needed to compile python packages
 # python3-dev gcc gcc-c++
 RUN dnf install -y \
-	NetworkManager openvswitch3.4 iproute python3 python3-devel openssh-clients gcc gcc-c++ \
+	NetworkManager openvswitch3.4 iproute python3 python3-devel openssh-clients gcc gcc-c++ --excludepkgs=iputils \
```
thom311, line 21: Hi. Is this really desirable?

Sure, `iputils` package may not be used in production, but it is a rather useful (and small?) tool for debugging (e.g. during `oc exec -ti pod/xxx -- /bin/bash`). Optimizing the size of containers may be a good idea, but did you measure how much this actually safes in size and whether it is worth it?

Also, IIC-763 is about CVEs (which?) in `iputils`. It's not clear which CVE exactly is this about, but the general solution is to fix the CVE in the affected package (by the package maintainers) and build a new container. Can you not just rebuild the container images?

Also, there really should be a justification in the commit message why things are done. OK, you refer to IIC-763, which is about CVEs. But could you replicate the high level summary of why this is done. IIC-763 is not accessible without login. There should be a discussion why things are done, so that in the future we can find the reasoning (e.g. when somebody wants to re-add the package).

saynb: > Optimizing the size of containers may be a good idea, but did you measure how much this actually safes in size and whether it is worth it? 

We do not care much about ping actually working in the VSP pod since we don't need ping to work there. `nmcli` is needed. So this seemed like a simple fix for the CVE. To just remove it. In production, we need to have ping utility in the host pods. The VSP pod (this dockerfile), doesn't lie in datapath at all

> Can you not just rebuild the container images?

I am not sure I understand the question. I am changing the Dockerfile in question. It will change the container image

swtaylor: This would be related to CVE-2025-47268 and we aren't specifically requesting an older version of iputils. Presumably the issue would go away by just regenerating the image?

bn222:
as of now, CI doesn't rebuild the changes that you're proposing here, so it's not testing the changes. We need to hash the dockerfile in it's entirety and then don't do an incremental build in case the hash changes. https://github.com/openshift/dpu-operator/blob/main/taskfiles/images.yaml#L26-L45

Dockerfile.IntelVSP.rhel
bn222: see my comment below

saynb:
> We need to hash the dockerfile in it's entirety and then don't do an incremental build in case the hash changes.

@bn222 I added an extra line in the dockerfile to force a rebuild. It worked locally for me so hopefully it will work here too. The solution that you are asking for here, that seems like keeping and maintaining computed hash of the dockerfile locally? Is that common practice? Seems like an involved change. How does the incremental vs full build get differentiated today?

bn222:
> @thom311 @bn222 What is exactly the further ask here? I have tested this change locally. Do you want to make permanent change to incremental build logic for this? If there is nothing else to do, can we please get an ok-to-test?

please fix the incremental build logic. It doesn't rebuild if you make a change to the dockerfile.

bn222:
First build is complete, any new build is incremental regardless of what the dockerfile contains. 
We lack differentiation based on file contents. That's what's missing.

PR #495, opened 2025-08-04
Intel ipu-opi-plugin upgrade

arunvelayutham:
Enable firewalld at ACC with restricted access (#460)

Enable firewalld at ACC with restricted access to allow only all the internal and p4 communication

NOTE: This must go in with CDA changes which handles all the cluster specific port enabling at firewalld along with primary network access.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/go-task/task/v3 v3.43.3
 	github.com/golang/glog v1.2.4
 	github.com/gorilla/mux v1.8.1
-	github.com/intel/ipu-opi-plugins/ipu-plugin v0.0.0-20250722202314-5c941620a662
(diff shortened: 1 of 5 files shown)


bn222:
The code now enables the firewall. This will probably not work as expected when running on OCP, only on microshift. Please check if that statement is correct. If so, then be inspired by https://github.com/openshift/dpu-operator/blob/44accd0d9ba5cf444a88e7d38825850ba914e220/internal/utils/cluster_environment.go#L61

arunvelayutham:
> The code now enables the firewall. This will probably not work as expected when running on OCP, only on microshift. Please check if that statement is correct. If so, then be inspired by
> 
> https://github.com/openshift/dpu-operator/blob/44accd0d9ba5cf444a88e7d38825850ba914e220/internal/utils/cluster_environment.go#L61

@bn222 please review the updated changes. I used ipuMode to ensure that the code runs only on the ACC

PR #498, opened 2025-08-11
CI for redfish adjustment

swtaylor:


swtaylor:
/hold
this run is for CI only. when successful will push it through the CDA workflow

swtaylor:
/test make-e2e-test
getting stuck after hugepages, which is completely irrelevant to the change i made :/

PR #500, opened 2025-08-12
[IIC-779] [th/bundle-fix-rbac-4.19] run `make bundle` to fix rbac permissions in release-4.19

thom311:
See https://issues.redhat.com/browse/IIC-779


bn222:
Can you add bundle-check target as well and add it as a CI lane?

PR #505, opened 2025-08-14
[th/add-images-in-env] config: fix missing environment variables for VSP images

thom311:
These entries are important to tell the operator which images to use.

Note that those quay.io images don't actually exist. These are placeholders only.

Also, the ART tooling needs to replace those values with the actual images. I don't actually understand how ART tooling mangles the manifest to replace those placeholds. If you know, please share the knowledge.

Also change "quay.io/openshift/origin-dpu-daemon:4.20" to "*:latest". If we tie this placeholder value 4.20, we would have to adjust it on new releases and during backports. For no reason at all. It's a placeholder only.

https://issues.redhat.com/browse/OCPBUGS-60512

### DIFF
--- a/bundle/manifests/dpu-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/dpu-operator.clusterserviceversion.yaml
@@ -416,6 +416,8 @@ spec:
                   value: quay.io/openshift/dpu-intel-ipu-p4sdk:latest
                 - name: MarvellVspImage
                   value: quay.io/openshift/dpu-marvell-vsp:latest
+                - name: MarvellVspCpAgentImage
+                  value: quay.io/openshift/dpu-marvell-cp-agent:latest
                 - name: IntelNetSecVspImage
                   value: quay.io/openshift/dpu-intel-netsec-vsp:latest
                 image: quay.io/openshift/dpu-operator:latest
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -86,6 +86,8 @@ spec:
           value: quay.io/openshift/dpu-intel-ipu-p4sdk:latest
         - name: MarvellVspImage
           value: quay.io/openshift/dpu-marvell-vsp:latest
+        - name: MarvellVspCpAgentImage
+          value: quay.io/openshift/dpu-marvell-cp-agent:latest
         - name: IntelNetSecVspImage
           value: quay.io/openshift/dpu-intel-netsec-vsp:latest
         #TODO: We will need pointers to all supported VSP images so we can dynamically detect the correct vendor to support at runtime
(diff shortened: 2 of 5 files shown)


bundle/manifests/dpu-operator.clusterserviceversion.yaml
```diff
@@ -364,9 +364,19 @@ spec:
                 - /manager
                 env:
                 - name: DpuOperatorDaemonImage
-                  value: quay.io/openshift/origin-dpu-daemon:4.20
```
wizhaoredhat, line 367: I think the ART team requires this to be 4:20. I would check with them first.

bundle/manifests/dpu-operator.clusterserviceversion.yaml
```diff
@@ -414,6 +414,10 @@ spec:
                   value: quay.io/openshift/dpu-intel-ipu-vsp:latest
                 - name: IntelVspP4Image
                   value: quay.io/openshift/dpu-intel-ipu-p4sdk:latest
+                - name: MarvellVspImage
+                  value: quay.io/openshift/dpu-marvell-vsp:latest
+                - name: IntelNetSecVspImage
```
wizhaoredhat, line 421: Ah I see it is there now.
Maybe we need a guide to onboard VSP images because there maybe more DPUs in the future that would need a VSP.

wizhaoredhat: Or a shell/python/make command.

wizhaoredhat: This comment shouldn't block this PR.

bundle/manifests/dpu-operator.clusterserviceversion.yaml
```diff
@@ -410,6 +410,10 @@ spec:
                   value: quay.io/openshift/dpu-daemon:latest
                 - name: NRIWebhookImage
                   value: quay.io/openshift/dpu-network-resources-injector:latest
+                - name: IntelVspImage
```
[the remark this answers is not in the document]
wizhaoredhat: Close this.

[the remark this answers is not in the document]
thom311: that comes in a later commit. The earlier commit (here) only adds the IPU images, those already exist in ART (and thus it's a different case from adding environment variables for images that don't exist at all).

if you prefer to not care about the individual commits, just look at the whole diff of the PR.

bundle/manifests/dpu-operator.clusterserviceversion.yaml
thom311: I think now that the exact URLs are there, it's gonna be easier.

Imagine you add a new environment variable. You cargo cult what is already and will grep for one of the existing URLs. Then you find all the locations and just add them there too.

It should be reasonably hard to miss for something that happens very infrequently.

Also, more important seems actual tests that the Operatorhub image work... those should also catch such a bug.

PR #506, opened 2025-08-14
Add DPU CR

bn222:
A long overdue PR. The in memory representation is wrapped into a runtime struct that also tracks which side managers are associated with the detected DPU. In addition, a status field tracks if init has been called correctly. The next PR will clean up the hack/setup.sh script. That script wasn't testing for correct init calls. That will become possible after this PR as the status field adds reporting capabilities to the operator. In the future, we will hook up more code to the status to track a disconnect (i.e. during reboot).

As part of the PR, we move to using the DPU CRs directly instead of first copying everything into another struct before copying it again.


bn222:
@wizhaoredhat ptal

internal/daemon/daemon_test.go
```diff
@@ -87,50 +114,75 @@ var _ = g.Describe("Full Daemon", func() {
...
 		g.It("Should have one DPU CR with IsDpuSide set", func() {
-			// Wait for DPU CR to be created
-			hostname, err := os.Hostname()
-			Expect(err).NotTo(HaveOccurred())
-
 			Eventually(func() error {
 				dpuCR := &v1.DataProcessingUnit{}
-				return k8sClient.Get(context.TODO(), client.ObjectKey{Name: hostname}, dpuCR)
-			}, 10*time.Second, 1*time.Second).Should(Succeed())
+				err := k8sClient.Get(context.TODO(), client.ObjectKey{Name: "intel-ipu"}, dpuCR)
```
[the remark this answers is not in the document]
bn222: This fake platform is used to simulate an environment with an IPU (Intelligence Processing Unit). By abstracting the platform at this high level, we avoid littering the codebase with specific mocks, which makes the code more robust and prevents it from becoming brittle. This approach ensures our tests are reliable and accurately represent the target hardware configuration.

api/v1/dataprocessingunit_types.go
```diff
@@ -31,7 +31,7 @@ type DataProcessingUnitSpec struct {
 	// DpuType is the type of DPU
 	DpuType string `json:"dpuType,omitempty"`
 	// IsDpuSide indicates if this DPU is on the DPU side
-	IsDpuSide bool `json:"isDpuSide,omitempty"`
```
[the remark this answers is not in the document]
bn222: This is more like a cosmetic change. If you fase it set to "false" which is seen as empty, then it doesn't explicitly show up as "false" in the ux. It just shows nothing in the IsDpuSide column.

internal/daemon/daemon_test.go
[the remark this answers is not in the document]
wizhaoredhat: Just don't understand why intel-ipu was chosen and why there isn't a Marvell DPU.

Aside: Isn't IPU = Infrastructure processing unit?

internal/daemon/daemon.go
```diff
@@ -213,12 +350,17 @@ func (d *Daemon) prepareCni() error {
...
+func (d *Daemon) runSideManager(mgr SideManager, identifier string, managerCtx context.Context) error {
 	err := mgr.StartVsp(managerCtx)
 	if err != nil {
 		d.log.Error(err, "Failed to start VSP in sideManager")
 		return err
 	}
+
+	// Mark the DPU as ready after successful VSP start
+	// TODO: hook this up to a health probe to the VSP so that it goes into NotReady while it doesn't respond
+	d.updateDpuStatus(identifier, "Ready")
```
wizhaoredhat, line 362: During the demo, you mentioned it would always be NotReady right now. Does this mean that this doesn't work? It looks fine to me.

internal/daemon/plugin/vendorplugin.go
```diff
@@ -109,6 +112,10 @@ func (g *GrpcPlugin) Start(ctx context.Context) (string, int32, error) {
 		ipPort, err := g.client.Init(ctx, &pb.InitRequest{DpuMode: g.dpuMode, DpuIdentifier: string(g.dpuIdentifier)})
 		if err != nil {
 			if strings.Contains(err.Error(), "already initialized") {
+				// VSP was already initialized, mark as initialized and return the error
+				g.initMutex.Lock()
+				g.initialized = false
```
wizhaoredhat, line 117: atomic.StoreInt32(&flag, 1)
and
val := atomic.LoadInt32(&flag)
might be better than adding a new mutex?

internal/platform/ipu.go
```diff
@@ -87,7 +87,9 @@ func (pi *IntelDetector) IsDpuPlatform(platform Platform) (bool, error) {
 func (pi *IntelDetector) GetDpuIdentifier(platform Platform, pci *ghw.PCIDevice) (plugin.DpuIdentifier, error) {
 	// TODO: rethink if it's possible to use something else than a pci address. Serial number doesn't seem to be the
 	// right choice for IPU.
-	identifier := fmt.Sprintf("intel-ipu-%s", pci.Address)
+	// Sanitize PCI address for Kubernetes resource name (replace : with -)
```
wizhaoredhat, line 90: Is this unique for IPU? Or can it be applied for Marvell?
For PTL we use the serial number.

wizhaoredhat:
overall lgtm, just some things not clear about.

internal/daemon/plugin/vendorplugin.go
bn222: - If this is a performance question, then we have to ask ourselves if saving a few nanoseconds in something that runs every second makes sense.
- If this is a readability/maintainablilty question, then we have to ask ourselves how easy it is to add more/make changes to an atomic int vs a line wrapped in a mutex.

It's easier to make changes if the code uses a mutex.

wizhaoredhat: The name of the mutex suggests that it is tied to the boolean. But I am ok with this.

Typically when mutexes are used in-line (not wrapped in a function), it makes the code harder to read.

In this case what makes the code easier to read is having a function SetInitDone(bool) where the mutex is inside the function.

internal/daemon/daemon_test.go
```diff
@@ -76,34 +114,70 @@ var _ = g.Describe("Full Daemon", func() {
...
 	})
 
 	g.Context("Running on a DPU", func() {
 		g.It("Should have one DPU CR with IsDpuSide set", func() {
-			// Eventually, it should show up as a CR
-			time.Sleep(5 * time.Second)
+			// First, wait for the DPU CR to be created
+			Eventually(func() error {
+				dpuCR := &v1.DataProcessingUnit{}
+				err := k8sClient.Get(context.TODO(), client.ObjectKey{Name: "intel-ipu"}, dpuCR)
```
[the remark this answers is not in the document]
wizhaoredhat: ok

internal/daemon/daemon.go
wizhaoredhat: ok makes sense.

PR #510, opened 2025-08-19
OCPBUGS-60658: [th/priviledged-sfc] Make the SFC container to be in privileged mode

thom311:
* Grant privileged access to SFC container
* Remove AllowPrivilegeEscalation since privileged automatically makes it true
* Correct typo in my-pod.yaml for privileged

Signed-off-by: Sayan Bandyopadhyay <sayan.bandyopadhyay@intel.com>
(cherry picked from commit 13b0558d64934eeabbf7488ce937376147adb4ca)

Without this, we don't have permissions to run the default "nginx" container. While we may revise this in the future, this is what we currently do on main branch. Backport.

### DIFF
--- a/examples/my-pod.yaml
+++ b/examples/my-pod.yaml
@@ -14,7 +14,7 @@ spec:
     command: ['/bin/sh', '-c', 'sleep infinity']
     imagePullPolicy: Always
     securityContext:
-      priviledged: true
+      privileged: true
       runAsNonRoot: false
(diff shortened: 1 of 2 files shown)


swtaylor:
/test make-e2e-test
on the last run, the worker node failed to install. i had manually run e2e-test directly on the host with no issue. this one looks like just a flake 
image

PR #517, opened 2025-08-25
vendor bump for dependencies

swtaylor:


thom311:
Could you please (in the git commit message) explain why things are done? Does this fix an actual problem, or do you just want to periodically pull latest versions. What is the backstory of this change.

It is interesting to know which problem this fixes (so when we encounter the problem, we know that this is the fix). If it does not fix any actual problem, that is also interesting to know (so I don't wonder which problem it fixes).

PR #524, opened 2025-08-27
Update ginkgo to prep for ipu bump

bn222:
Pull in new gingko for making the delta of smaller between this repo and ipu plugin repo. There is a small change in how parallelism is handled, so this bump in isolation will show if our tests were written in a robust enough way.


bn222:
already passed. just a retest after a merge.

PR #530, opened 2025-08-29
Remove Mode field and Add support for Layer 3 Network Functions on PTL

wizhaoredhat:
Remove Mode field for proper "auto" support

-  Remove the Mode field in the DPU Operator Config CRD because it is no longer needed when we auto detect whether each node is running with vs. on a DPU. This auto detection was completed in prior commits.
-  Remove all references to the Mode field in tests and code
- When creating NADs, we create both NADs for host and DPU
- This does not break existing setups since the auto detection code is in and the mode field is only used to control the creation of NADs.

Add support for Layer 3 Network Functions
    

- Revised VethPairs data structure to handle lookups by MAC address. This is needed since the API for creating network functions is by MAC address of the veth interface.
- To support Layer 3 Network Functions, the veth ports are simply connected to the OVS bridge. Pods on the host can reach other pods without going through the network function. However Pods on the host can reach the network function in the same subnet meaning that the network function must assign an IP address on the veth interfaces.
- Currently no OvS rules are needed to support Layer 3 Network Functions

### DIFF
--- a/api/v1/dpuoperatorconfig_types.go
+++ b/api/v1/dpuoperatorconfig_types.go
@@ -27,10 +27,6 @@ import (
 
 // DpuOperatorConfigSpec defines the desired state of DpuOperatorConfig
 type DpuOperatorConfigSpec struct {
-	// Mode can be "host" or "dpu" and it defines on which side we are
-	// TODO: add support for auto
-	Mode string `json:"mode,omitempty"`
-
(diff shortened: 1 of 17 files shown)


wizhaoredhat:
@bn222 / @thom311  PTAL

wizhaoredhat:
Passed prior to trying to solve the verified label issue.

PR #532, opened 2025-09-03
[th/marvell-boot-live-iso-dpu] marvell: boot live ISO for accessing worker node

thom311:
We use pxeboot to install RHEL on the DPU. For that, we need to SSH into
the host (that has the DPU inside) and run a podman command. That
requires that the host is up and accessible via SSH.
    
If the host is not accessible, then fallback to boot a CoreOS Live iso.
   
https://issues.redhat.com/browse/MDC-108
https://issues.redhat.com/browse/IIC-806


thom311:
waiting on https://github.com/bn222/cluster-deployment-automation/pull/383

PR #535, opened 2025-09-04
dpuCR data structure should not contain metadata

wizhaoredhat:
dpuCR is used for CR creation and should remain clean of metadata.


wizhaoredhat:
@bn222 PTAL. Not sure if you want to merge this in then update the split brain issue.

PR #537, opened 2025-09-05
Simplify scheme handling

bn222:


bn222:

Marvell will pass as well since this is a vendor agnostic change that behaves uniformly across vendors.

Already passed.

/verified bypass

PR #542, opened 2025-09-08
Revise SFC design and introduce dpuside node labels

wizhaoredhat:


internal/daemon/dpusidemanager.go
```diff
@@ -261,9 +263,20 @@ func (d *DpuSideManager) setupReconcilers() {
...
 			NewCache: func(config *rest.Config, opts cache.Options) (cache.Cache, error) {
+				// Watch ServiceFunctionChains only in operator namespace, but Pods in both namespaces.
+				// ServiceFunctionChain CR is namespace scoped to be tied to the dpu-operator namespace.
+				// It is not a clusterwide resource.
 				opts.DefaultNamespaces = map[string]cache.Config{
 					vars.Namespace: {},
 				}
+				opts.ByObject = map[client.Object]cache.ByObject{
+					&corev1.Pod{}: {
+						Namespaces: map[string]cache.Config{
```
bn222, line 274: This is a bit weird

internal/daemon/hostsidemanager.go
```diff
@@ -323,9 +325,18 @@ func (d *HostSideManager) setupReconcilers() {
 		mgr, err := ctrl.NewManager(d.config, ctrl.Options{
 			Scheme: scheme.Scheme,
 			NewCache: func(config *rest.Config, opts cache.Options) (cache.Cache, error) {
+				// Watch ServiceFunctionChains only in operator namespace, but Pods in both namespaces
 				opts.DefaultNamespaces = map[string]cache.Config{
 					vars.Namespace: {},
 				}
+				opts.ByObject = map[client.Object]cache.ByObject{
+					&corev1.Pod{}: {
+						Namespaces: map[string]cache.Config{
```
bn222, line 334: Same. There should be a cleaner way to set this

internal/daemon/sfc-reconciler/sfc.go
```diff
@@ -171,6 +207,47 @@ func (r *SfcReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+			if err := r.cleanupNetworkFunctionPods(ctx, sfc); err != nil {
+				r.log.Error(err, "Failed to cleanup network function pods")
+				return ctrl.Result{RequeueAfter: time.Minute}, err
+			}
+		}
+
+		// Remove our finalizer
+		if controllerutil.ContainsFinalizer(sfc, SfcFinalizerName) {
+			controllerutil.RemoveFinalizer(sfc, SfcFinalizerName)
+			if err := r.Update(ctx, sfc); err != nil {
```
bn222, line 232: if inside if inside if inside if.

Can we refactor? At least move all of the finalizer stuff into a function

internal/controller/bindata/daemon/04.daemon_cluster_role.yaml
```diff
@@ -19,6 +19,7 @@ rules:
   - get
   - list
   - watch
+  - update
```
bn222, line 22: We should wait until the round-trip PR goes in because there's a substantial change on how the rbac is done there.

internal/daemon/daemon.go
```diff
@@ -419,3 +427,59 @@ func (d *Daemon) runSideManager(mgr SideManager, identifier string, managerCtx c
 	}
 	return nil
 }
+
+// updateNodeLabels updates the node labels based on the detected DPUs. This is done independently since labels can be
+// removed or modified accidently. We always want this label to be set/deleted properly.
+func (d *Daemon) updateNodeLabels() error {
```
bn222, line 433: We're going back to labels instead of DPU CR? Not clear why

examples/my-pod.yaml
```diff
@@ -7,9 +7,8 @@ metadata:
     k8s.v1.cni.cncf.io/networks: default-sriov-net
 spec:
   nodeSelector:
-    # TODO: when annotation for DPU vs. Host is added, we can use the proper annotation here.
     # kubernetes.io/hostname: worker-229
-    dpu: "true"
+    dpu.config.openshift.io/dpuside: "dpu-host"
```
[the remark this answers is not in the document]
wizhaoredhat: I feel like with the DPU Operator it is all in one solution. We don't rely on external components. So NFD is probably useless to us. I.e. from my understanding DPF operator does not have detection and relies on NFD to provide that detection logic.

internal/daemon/daemon.go
wizhaoredhat: You can't use the DPU CR as a node selector. So we need both to be able to schedule nodes properly in a one cluster deployment.

internal/daemon/dpusidemanager.go
wizhaoredhat: Could you please elaborate on the weirdness?

internal/controller/bindata/daemon/04.daemon_cluster_role.yaml
wizhaoredhat: It should be fine to merge in since it is appending to the end of the list.

internal/daemon/sfc-reconciler/sfc.go
wizhaoredhat: I can add a function. The subsequent commit adds the rbac.

examples/my-pod.yaml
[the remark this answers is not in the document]
bn222: nfd adds a label, we need a way to specifiy which label to watch for. We don't need to use nfd, but allowing the user to specify what key-value to use is useful *if* they want to use it in conjunction with nfd. IIC-906 speaks about providing that optionality. 

Does your proposed approach work in case the user decides to use nfd?

examples/sfc.yaml
```diff
@@ -2,8 +2,11 @@ apiVersion: config.openshift.io/v1
 kind: ServiceFunctionChain
 metadata:
   name: sfc-name
-  namespace: default
+  namespace: openshift-dpu-operator
```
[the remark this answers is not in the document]
wizhaoredhat: my mistake on the first commit. Fixed.

examples/my-pod.yaml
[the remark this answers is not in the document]
wizhaoredhat: Why would the user need to use NFD? I don't understand what is meant here.

internal/controller/bindata/daemon/04.daemon_cluster_role.yaml
wizhaoredhat: :sweat_smile: Someone needs to do the rebase, it's either going to be done in this PR which touches rbac, or yours which touches rbac.

internal/daemon/dpusidemanager.go
bn222: Now I see what you're doing. You're configuring the informer to only cache local SFCs?

internal/daemon/daemon.go
wizhaoredhat: This is done in code. Which you can take values from any resource and apply to any other resource. But this doesn't allow you to create workloads that you want to run on either the hosts or the dpu externally.

This is for workload pods that the user creates and that we have no control over.

wizhaoredhat:
Closing this PR. Chaning the SFC yaml to do this is too disruptive. If we need to fix up the SFC yaml, we can revist this PR.

PR #546, opened 2025-09-10
Add DPU config labels to identify nodes with DPU or DPU Hosts

wizhaoredhat:
Resolve v1 name conflicts in the imports for daemon.go

These node labels are useful to schedule workloads on DPUs or DPU Hosts. This is not a replacement of the "dpu: true" label.


wizhaoredhat:
@bn222 PTAL

PR #548, opened 2025-09-10
Use mutation pattern instead when creating NF pod with SFC yaml

wizhaoredhat:
We shouldn't directly use the desired NF pod defition when attempting to update pods we create with the SFC yaml. This will cause issues in the metadata fields leading to errors such as "the object has been modified; please apply your changes to the latest version and try again" when updating with the same NF pod definition.


wizhaoredhat:
Doesn't look like an issue anymore.

PR #553, opened 2025-09-15
Add status field to DPU Operator Config CR

bn222:
This change allows the standard wait command to be used for reconciliation. This is a significant improvement over manually inspecting logs or relying on crude `sleep` commands, which are unreliable and difficult to debug. Tests also check for errors in the reconciliation now.

This provides a more robust and idiomatic way to wait for reconciliation from the CLI. A follow-up commit will extend this functionality to also wait on a DPU CR.

Fix unifying the yaml file while at it as well.


bn222:
@wizhaoredhat ptal

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -106,26 +114,97 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		r.setNotReadyCondition(dpuOperatorConfig, reason, firstError.err.Error())
+		if updateErr := r.Status().Update(ctx, dpuOperatorConfig); updateErr != nil {
+			logger.Error(updateErr, "Failed to update status with error condition")
+		}
+		return firstError.err
+	}
+
+	// All components reconciled successfully - set Ready condition
+	r.setReadyCondition(dpuOperatorConfig, "AllComponentsReady", "DPU Daemons and Network Functions are ready")
+	if updateErr := r.Status().Update(ctx, dpuOperatorConfig); updateErr != nil {
```
wizhaoredhat, line 200: I think we should "get" before update. Or use the mutation design. I think tracing the code you do multiple updates to the CR. The reason why this works is because the data structure is updated when "update" is called. But if "dpuOperatorConfig" is ever updated by a daemon then this breaks. You have a mutual exclusion issue.
Which is when you mentioned about one field belonging to the daemon and 1 field belonging to the controller. That it is not as easy as we picture since "update" is not really a merge and if it detects changes that it has not included it will fail.

bn222: Get is on this line https://github.com/openshift/dpu-operator/pull/553/files/a04da8f6b5d746f93b26c861a7d223a25e021a59#diff-ac97373df236b7f27a14df2d71cf3b436098b60ff6ae8fccdb17c6a6266a33ecR108

If we would do a get just before the update, then that would only narrow the distance between the get and the update.

The daemon doesn't touch this. If it would, the update here would fail because resourceVersion would not be the same and we would error out on line https://github.com/openshift/dpu-operator/pull/553/files/a04da8f6b5d746f93b26c861a7d223a25e021a59#diff-ac97373df236b7f27a14df2d71cf3b436098b60ff6ae8fccdb17c6a6266a33ecR143. That's how k8s resolves the conflict (which doesn't exist here now, and probably will not exist ever for DpuOperatorConfig CR).

internal/testutils/testcluster.go
```diff
@@ -493,6 +495,23 @@ func CreateDpuOperatorCR(client client.Client, cr *configv1.DpuOperatorConfig) {
 	Eventually(func() error {
 		return client.Get(context.Background(), types.NamespacedName{Namespace: cr.GetNamespace(), Name: cr.GetName()}, &found)
 	}, TestAPITimeout, TestRetryInterval).Should(Succeed())
+
+	// Set to Ready state for testing - avoids waiting for real reconciliation
```
wizhaoredhat, line 499: I don't understand this part...

It seems you want to over ride what the controller is doing setting the status?
Wouldn't this not test the reconcile loop code?

bn222: Some test don't have a controller. It's the controller that sets it to true. For the sake of those tests, we put the component under test into an env where the CR has a ready state.

internal/controller/dpuoperatorconfig_controller.go
wizhaoredhat: As long as the daemons never touch the DPU Config CR, this should be good.

However this is a note for when we update the DPU CR modifications you plan to do.

internal/controller/dpuoperatorconfig_controller.go
```diff
@@ -106,26 +116,90 @@ func (r *DpuOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+func (r *DpuOperatorConfigReconciler) updateStatus(ctx context.Context, dpuOperatorConfig *configv1.DpuOperatorConfig, reconcileErrors []componentError) error {
+	logger := log.FromContext(ctx)
+
+	if len(reconcileErrors) > 0 {
+		// Set NotReady condition with the first error, including component info in reason
+		firstError := reconcileErrors[0]
+		reason := fmt.Sprintf("%sError", firstError.component)
+		r.setReadyCondition(dpuOperatorConfig, metav1.ConditionFalse, reason, firstError.err.Error())
+		if updateErr := r.Status().Update(ctx, dpuOperatorConfig); updateErr != nil {
+			logger.Error(updateErr, "Failed to update status with error condition")
```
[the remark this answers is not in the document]
bn222: Arguably, both are important. If we look at the logs, we probably want know the root cause first which is firstError. If we fix that, it might resolve the whole problem. If we fail to update, we need to return that instead. I'll make a change to handle both paths (update failed while trying to set the first error, and update succeeded buy have a first error to return).

PR #562, opened 2025-09-19
Marvell VSP: Add configurable MAC learning for OVS data plane

alkama-hasan:
Added isMacLearning parameter to control bridge switching behavior. When true, configures highest priority NORMAL flow for MAC-based learning. When false, allows traffic routing through network functions.


alkama-hasan:
>please rebase the PR to latest main branch.
Done.
> let's see how the tests work with SKIP_NF_TESTING removed to see how all tests would pass.
By default it will run a full test-suite? if yes we can definetly try or do you think any changes in e2e-test is needed ?

thom311:
> By default it will run a full test-suite? if yes we can definetly try or do you think any changes in e2e-test is needed ?

If we remove all occurrances of the `skipNetworkFunctionTesting` variable, then the tests will run.

In the best case, the test then just passes and we are good. If the test then fails, we will need to investigate (and maybe fix later, that test failure does not have to block this PR).

In step 1, let's see how all the tests would work.

alkama-hasan:
@thom311 
>You didn't yet add a commit to drop skipNetworkFunctionTesting. So the most interesting tests will be skipped.
Anyway.

as i see all the test passed. i just wanted to know what is the difference between 
make-e2e-test 
and make-e2e-test-marvell?

bn222:
Instead of moving things into jenkins again, we should determine that external interface dynamically or we should move the CI to 2 workers with a Marvell DPU in each.

thom311:
let's do https://github.com/openshift/dpu-operator/pull/579.

And, set `EXTERNAL_CLIENT_DEV=eno2` in Jenkins, for the Marvell job.

PR #565, opened 2025-09-19
Compare each field in the DPU CR's status field for changes

wizhaoredhat:
Prior to this commit, the DPU CR when using deep equal on status would flood logs with updates. However these updates contain no actual changes. The reason is due to using deep equal on a data structure containing metadata from k8s. In this case it would be LastTransitionTime.
    
Instead we need to compare each relevant field in the Condition under the status field. This stops the flooding of updates and the logs for the update.


internal/daemon/daemon.go
```diff
@@ -355,6 +357,47 @@ func (d *Daemon) syncSingleDpuCR(dpuCR *configv1.DataProcessingUnit, existingCRM
...
+		if currentCondition.Status != desiredCondition.Status ||
+			currentCondition.Reason != desiredCondition.Reason ||
+			currentCondition.Message != desiredCondition.Message ||
+			currentCondition.ObservedGeneration != desiredCondition.ObservedGeneration {
+			return true
+		}
+	}
+
+	// Check if any existing conditions are not in desired (i.e., should be removed)
+	for _, currentCondition := range current {
```
thom311, line 384: why is this loop here exactly?

At the beginning, we check `if len(current) != len(desired)`.
Then, the loop
```go
    for _, desiredCondition := range desired {
		currentCondition := meta.FindStatusCondition(current, desiredCondition.Type)
```
will be sufficient if-and-only-if we are guaranteed that at least one of current or desired only contains **unique** types.

If both lists possibly contain duplicated types, then the code so far is not sufficient. But in that case, the addtional loop here doesn't fix that either...

wizhaoredhat: Understood, great catch. I fixed it.

PR #572, opened 2025-09-24
Fix the cleanup of tests by blocking when deleting the CR

bn222:
Block on removal of CR so that next test doesn't fail due to leak

### DIFF
--- a/internal/testutils/testcluster.go
+++ b/internal/testutils/testcluster.go
@@ -683,35 +683,60 @@ func EventuallyDpuOperatorConfigReady(c client.Client, logger logr.Logger, cr *c
 	return dpuOperatorConfig
 }
 
-func EventuallyDpuOperatorConfigDeleted(c client.Client, name string, namespace string, timeout time.Duration, interval time.Duration) {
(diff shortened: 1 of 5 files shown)


bn222:
@thom311 ptal again, forgot to go mod tidy in the api dir to pull in helpers

internal/testutils/testcluster.go
```diff
@@ -683,35 +683,64 @@ func EventuallyDpuOperatorConfigReady(c client.Client, logger logr.Logger, cr *c
...
-				return nil
-			}
+			crList := &configv1.DpuOperatorConfigList{}
+			err := c.List(context.Background(), crList)
 			if err != nil {
-				return fmt.Errorf("error checking DpuOperatorConfig %s in %s: %v", name, namespace, err)
+				return fmt.Errorf("failed to list DpuOperatorConfigs: %v", err)
+			}
+
+			if len(crList.Items) > 0 {
```
wizhaoredhat, line 720: Duplication of come from lines 694-706. We are just checking the finalizer label for the CR right? And reporting it?

PR #573, opened 2025-09-24
fix bundle and add bundle check so that we don't forget anymore

bn222:


Makefile
```diff
@@ -222,22 +222,6 @@ bundle:
 bundle-check:
 	go run tools/task/task.go bundle-check
 
-.PHONY: prow-ci-bundle-check
```
wizhaoredhat, line 225: Why is this removed? We have a prow job that does this, no?

Please update commit messages and PR description as to why. E.g. was this ever properly working before? yes/no? Why was it moved to taskfile from Makefile. It seems prow-ci-bundle-check existed, what happend?

PR #575, opened 2025-09-24
Handle the case where IPU takes a longer time to come up

bn222:
Edge case when ipu reboots that wait will fail since microshift will go down. Instead, wait on the host side, and then wait for the DPU side. If the host is ready, the DPU should be too.

This edge case will be better taken care of when we add IIC-960. That will only allow the host side DPU CR to go into ready state if the DPU responded to at least 1 health probe which it will only do if the init on the dpu also succeeded.

### DIFF
--- a/hack/setup.sh
+++ b/hack/setup.sh
@@ -34,14 +34,15 @@ KUBECONFIG=/root/kubeconfig.microshift oc label nodes --all dpu=true
 KUBECONFIG=/root/kubeconfig.microshift oc create -f examples/config.yaml
 KUBECONFIG=/root/kubeconfig.ocpcluster oc create -f examples/config.yaml
 
-# Wait for DpuOperatorConfig to be Ready on both clusters
-echo "Waiting for DpuOperatorConfig to be Ready on microshift cluster..."
-KUBECONFIG=/root/kubeconfig.microshift oc wait --for=condition=Ready dpuoperatorconfig/dpu-operator-config -n openshift-dpu-operator --timeout=1m
 echo "Waiting for DpuOperatorConfig to be Ready on OCP cluster..."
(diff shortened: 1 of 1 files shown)


bn222:
@wizhaoredhat ptal

hack/setup.sh
```diff
@@ -34,14 +34,15 @@ KUBECONFIG=/root/kubeconfig.microshift oc label nodes --all dpu=true
...
 
-# Wait for DpuOperatorConfig to be Ready on both clusters
-echo "Waiting for DpuOperatorConfig to be Ready on microshift cluster..."
-KUBECONFIG=/root/kubeconfig.microshift oc wait --for=condition=Ready dpuoperatorconfig/dpu-operator-config -n openshift-dpu-operator --timeout=1m
 echo "Waiting for DpuOperatorConfig to be Ready on OCP cluster..."
 KUBECONFIG=/root/kubeconfig.ocpcluster oc wait --for=condition=Ready dpuoperatorconfig/dpu-operator-config -n openshift-dpu-operator --timeout=1m
+echo "Waiting for DPU to be Ready on OCP cluster..."
+wait_for_dpu /root/kubeconfig.ocpcluster
 
+echo "Waiting for DpuOperatorConfig to be Ready on microshift cluster..."
```
wizhaoredhat, line 42: Commit message should explain why this was an issue. And why moving it helps.

wizhaoredhat: E.g. explain why this shuffling solves the problem.
Not just the effect of the changes. Explanations help reviewers understand and people that look at the history of the code to understand what happend.

bn222:
I mentioned "If the host is ready, the DPU should be too." By shuffling we wait on the host first, then on the DPU.

PR #578, opened 2025-09-25
Heart beat

bn222:
Ping from the host to the card to report the health on the host and on the card in the status of the DPU CR.

- If the card goes down, the status reflects that on the host
- If the host goes down, the status reflects that on the card
- When the card is coming up, it's not enough anymore to simply call init. The connection between the host and dpu side should be up.
- If a ping failed for some time, the status will be moved back to false.

This makes the whole solution more robust and makes it so that hack/setup.sh doesn't need to wait on both DPU CRs in a specific order. We might even remove the wait on one of the DPU CRs since the status is now available on both sides.

### DIFF
--- a/internal/daemon/hostsidemanager.go
+++ b/internal/daemon/hostsidemanager.go
@@ -221,6 +221,18 @@ func (d *HostSideManager) cniCmdDelHandler(req *cnitypes.PodRequest) (*cni100.Re
 	return nil, nil
 }
 
+func (d *HostSideManager) setPing(ping time.Time) {
+	d.pingMutex.Lock()
+	d.lastSuccessfulPing = ping
+	d.pingMutex.Unlock()
+}
+
+func (d *HostSideManager) getPing() time.Time {
+	d.pingMutex.RLock()
+	defer d.pingMutex.RUnlock()
+	return d.lastSuccessfulPing
+}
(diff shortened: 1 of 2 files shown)


bn222:
Demo:

```
# oc get dpu -w
NAME                     DPU PRODUCT       DPU SIDE   NODE NAME    STATUS
intel-ipu-0000-06-00.0   Intel IPU E2100   false      worker-215   True
intel-ipu-0000-06-00.0   Intel IPU E2100   false      worker-215   False
```

When a DPU goes down, the status is set to False

internal/daemon/hostsidemanager.go
```diff
@@ -296,6 +356,24 @@ func (d *HostSideManager) Serve(ctx context.Context, listener net.Listener) erro
...
 
+	wg.Add(1)
+	go func() {
+		d.log.Info("Starting ping client")
+		ticker := time.NewTicker(1 * time.Second)
+		defer ticker.Stop()
+		defer wg.Done()
+
+		for {
+			select {
```
wizhaoredhat, line 367: Can we have a pattern similar to the above like starting manager and starting device plugin server?

bn222: done

PR #585, opened 2025-09-30
Move vsp-p4 plugin yaml context one level above

arunvelayutham:
Move vsp-p4 plugin yaml context one level above.


arunvelayutham:
Please hold on from reviewing this PR until I raise intel VSP plugin PR which will handle the P4 rule clean-up by Finalizer or preStop approach.

PR #586, opened 2025-09-30
Refactor 2 cluster in E2E tests and add 1 cluster support

wizhaoredhat:
The current e2e tests assumes 2 clusters. Depending on which kubeconfigs exists we should detect whether it is in 2 cluster or 1 cluster modes. Once the 2 vs 1 cluster deployment is determined, there are code changes to make sure the tests work in either approach.

### DIFF
--- a/hack/traffic_flow_tests.sh
+++ b/hack/traffic_flow_tests.sh
@@ -15,14 +15,6 @@ if [ -z "$worker" ]; then
   exit 1
 fi
 
-export KUBECONFIG=/root/kubeconfig.microshift
-nodes=$(oc get nodes)
-export acc=$(echo "$nodes" | grep -oP '^\d+-acc')
-if [ -z "$acc" ]; then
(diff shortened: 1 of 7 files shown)


wizhaoredhat:
@bn222 PTAL

hack/traffic_flow_tests.sh
```diff
@@ -15,14 +15,6 @@ if [ -z "$worker" ]; then
...
 fi
 
-export KUBECONFIG=/root/kubeconfig.microshift
-nodes=$(oc get nodes)
-export acc=$(echo "$nodes" | grep -oP '^\d+-acc')
-if [ -z "$acc" ]; then
-  echo "Error: acc is empty"
-  exit 1
-fi
-
```
bn222, line 25: Nice catch. This is now taken care of by the setup.sh script and indeed we don't need this anymore

PR #588, opened 2025-10-02
Clean up cp agent build process

bn222:


internal/controller/bindata/vsp/marvell-dpu/99.vsp-pod.yaml
```diff
@@ -24,6 +24,9 @@ spec:
       runAsUser: 0
     command: ["/vsp-mrvl"]
     args: []
+    env:
+    - name: MarvellVspCpAgentImage
+      value: "{{.MarvellVspCpAgentImage}}"
```
thom311, line 29: variable also needs to go to `config/dev/local-images-template.yaml`

bn222:
IPU in recovery mode. Will be fixed separately. Passed on Marvell, which is the most relevant for the test. IPU test isn't affected by this.

PR #589, opened 2025-10-02
Add redeploy task for 1 cluster

wizhaoredhat:
All the original tasks are expecting 2 clusters hense using 2 kubeconfigs. However for 1 cluster we need a duplicate set to only look for 1 kubeconfig (OCP).


wizhaoredhat:
Save some CI bandwidth for other PRs. This change doesn't affect 2 cluster.

PR #591, opened 2025-10-03
Move Intel vsp-p4 plugin yaml context one level above

arunvelayutham:
Move Intel vsp-p4 plugin yaml context one level above


internal/controller/bindata/vsp/intel-ipu/04.vsp_p4_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: vsp-p4-role-binding
+  namespace: {{.Namespace}}
+subjects:
+- kind: ServiceAccount
+  name: vsp-p4-sa
+roleRef:
+  kind: Role
+  name: vsp-p4-role
+  apiGroup: rbac.authorization.k8s.io
```
thom311, line 12: the commit takes the files from vendor/github.com/intel/ipu-opi-plugins/ipu-plugin/pkg/infrapod/bindata/vsp-p4-pvc.

For one, the commit message should explain where exactly those files come from. Also, it is not clear why this is done, why it is necessary or why it is a good idea.

Also, the commit renames the files. I understand you try to renumber them, but is that really necessary? Even if you end up with two files starting with `01.`, does it actually hurt? As the files are literally identical as from the vendored directory, can we keep the filenames unchanged?

PR #673, opened 2026-08-17
feat(api): allow ServiceFunctionChain network functions to specify a Helm chart

kartikeyg0104:
## What

Add an optional Helm chart workload source on `ServiceFunctionChain.spec.networkFunctions[]`, so a network function can be described either as a container image or as a Helm chart.

```go
type NetworkFunction struct {
    Name  string            `json:"name"`
    Image string            `json:"image,omitempty"`
    Chart *HelmChartSource  `json:"chart,omitempty"`
}

type HelmChartSource struct {
    Repository string `json:"repository"`
    Name       string `json:"name"`
    Version    string `json:"version"`
}
```

`Image` is now optional. A CEL rule on the CRD requires exactly one of `image` or `chart`.

Generated with `make generate` and `make manifests`. Vendored API copies updated so `check-vendor` stays green.

## Why

`NetworkFunction` currently only accepts a raw container image. That is enough for the Intel and Marvell vendor plugins, which deploy a pod from `nf.Image`.

NVIDIA DPF has no field that accepts a bare image. A `DPUService` requires `helmChart.source.repoURL` and `version`. Without a chart coordinate on the OPI type, an NVIDIA VSP cannot express a DPF-backed network function in a `ServiceFunctionChain` without inventing a private mapping outside the API.

This is an additive, backwards-compatible CRD change: existing manifests that set `image` continue to validate.

## Compatibility

- Existing `image:`-only ServiceFunctionChain objects remain valid.
- New objects may set `chart:` instead of `image:`.
- Objects that set both, or neither, are rejected by CEL: `exactly one of image or chart must be set`.
- `HelmChartSource.{repository,name,version}` are all required when `chart` is present.

## Follow-up (not in this PR)

The SFC reconciler still creates a pod from `nf.Image` in `internal/daemon/sfc-reconciler/sfc.go`. Chart-backed network functions will be consumed by vendor-specific plugins (the NVIDIA VSP in particular). Controller/VSP handling of `Chart` should land in a subsequent PR so this API change can merge independently.

`bundle/` and `manifests/stable/` were not regenerated; `make manifests` only writes `config/crd/bases`. Happy to run `make bundle` if reviewers want those copies updated in this PR.

## Example

```yaml
apiVersion: config.openshift.io/v1
kind: ServiceFunctionChain
metadata:
  name: nvidia-hbn
spec:
  networkFunctions:
    - name: hbn
      chart:
        repository: https://helm.ngc.nvidia.com/nvidia/doca
        name: dpf-dpu-service
        version: v25.10.1
```


wizhaoredhat:
Hi @kartikeyg0104 please create this PR in the OPI fork of the DPU Operator. This is now downstream specific for OpenShift.

kartikeyg0104:
Thanks @wizhaoredhat — that matches the governance we needed.

`openshift/dpu-operator` is the OpenShift downstream fork. The Helm-chart `NetworkFunction` API belongs upstream in `opiproject/dpu-operator` so NVIDIA (and other vendors) can consume it without OpenShift-specific coupling.

Closing this PR and opening the same change against `opiproject/dpu-operator`.

